PR #1, opened 2016-02-09
Signing request with HMAC-SHA256

erwinfoekema:
Signing request with HMAC-SHA256


mtoneman:
Hi @ErwinFo , thanks for the work on this. I just noticed that you hardcoded the session validity.

`var sessionValidity : String 		= "2019-04-06T18:10:29+02:00";
`
This is dangerous, as you are creating a session with a very long validity. It would be better to set this dynamically to something like one hour in the future so it can't be abused (a second issue is of course that all payments will magically start failing 3 years from now).

ErwinFo:
Hi @mtoneman I completely agree with you, I will make some adjustments to hardcoded variables when I finish with the other files that need changes in order to work with SHA256.

However, if you look at the current getPaymentMethods.ds file, you can see that 
var sessionValidity : String = "2019-04-06T18:10:29+02:00"; was allready hardcoded.I kept it this way.

There allot more variables hardcoded like URL's (they could also change in the future)
And in the file https://github.com/Adyen/demandware/blob/master/cartridges/int_adyen/cartridge/scripts/adyenRedirectVerification.ds

the sessionValidity is also hardcoded.

SqrTT:
Hi @erwinfoekema, excellent work. I've used your changes on real project and it works fine.

erwinfoekema:
Hi @SqrTT thank you, and thanks for confirming and testing it.

rikterbeek:
hi @erwinfoekema,

Many thanks for this pull request I have merged it into our codebase.

Regards,
Rik
Adyen

PR #16, opened 2017-04-05
Fixing bad references of site preferences and param map

shr1k4nt:
Fixing bad references of two things:
1. CurrentHttpParameterMap should be request.httpParametermap
2. Updating the site preference ids to match the corrent ones as coming up in the metadata xml file.


rikterbeek:
Hi @shr1k4nt,

Many thanks for this PR! I will merge it into our codebase.

Regards,
Rik
Adyen

PR #39, opened 2017-08-03
ShipBeforeDate configurable

maassenbas:
ShipBeforeDate in Custom Preferences

### DIFF
--- a/cartridges/bm_adyen/site-template/meta/system-objecttype-extensions.xml
+++ b/cartridges/bm_adyen/site-template/meta/system-objecttype-extensions.xml
@@ -241,6 +241,14 @@
                 <externally-managed-flag>false</externally-managed-flag>
                 <min-length>0</min-length>
             </attribute-definition>
+            <attribute-definition attribute-id="AdyenShipBeforeDate">
+                <display-name xml:lang="x-default">ShipBeforeDate</display-name>
+                <description xml:lang="x-default">Number of days before shipping</description>
+                <type>integer</type>
+                <mandatory-flag>false</mandatory-flag>
+                <externally-managed-flag>false</externally-managed-flag>
+                <default-value>5</default-value>
+            </attribute-definition>
             <attribute-definition attribute-id="AdyenPaymentMethodsServicePrefix">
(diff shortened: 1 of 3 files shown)


cartridges/bm_adyen/site-template/meta/system-objecttype-extensions.xml
```diff
@@ -379,6 +387,7 @@
                 <attribute attribute-id="Adyen_HMACkey"/>          
                 <attribute attribute-id="Adyen_Mode"/>
                 <attribute attribute-id="Adyen_Debug"/>
+                <attribute attribute-id="AdyenShipBeforeDate"/>
```
rikterbeek: Could you add in the configuration the value 5 as default so it is always filled in and no error in request are made

cartridges/bm_adyen/site-template/meta/system-objecttype-extensions.xml
```diff
@@ -241,6 +241,14 @@
                 <externally-managed-flag>false</externally-managed-flag>
                 <min-length>0</min-length>
             </attribute-definition>
+            <attribute-definition attribute-id="AdyenShipBeforeDate">
```
rikterbeek: could you change this to AdyenShipBeforeDays rather then AdyenShipBeforeDate because it does not contain a data

cartridges/int_adyen/cartridge/scripts/adyenRedirectVerificationSHA256.ds
```diff
@@ -108,8 +108,9 @@ function verify (args) {
 	sessionValidity.setHours(sessionValidity.getHours() + 1);
 	args.sessionValidity = sessionValidity.toISOString();
 
+ 	var shipBeforeDays = Site.getCurrent().getCustomPreferenceValue("AdyenShipBeforeDate");
     var shipBeforeDate : String = new Date();
-    shipBeforeDate.setDate(shipBeforeDate.getDate() + 5);
+    shipBeforeDate.setDate(shipBeforeDate.getDate() + shipBeforeDays);
```
rikterbeek: if the value is not set could you fallback to 5 so the payment does not fail ?

PR #54, opened 2018-04-11
consistency marketplace and github

maassenbas:
Deleted/renamed files


rikterbeek:
@maassenbas let's keep the meta data in version control so we can keep track of changes.

PR #74, opened 2018-08-14
Recurring HPP

maassenbas:
Check if Recurring is enabled and customer is logged in for HPP Redirect


rikterbeek:
For now the quickfix is oke, but I don't like the setting AdyenRecurringPaymentsEnabled we should think to do this better. If you want oneclick you need to enable AdyenRecurringPaymentsEnabled and put type on oneclick not very user friendly.

PR #82, opened 2018-09-12
Release consistency

maassenbas:
Making the release 18.1.2 on LINK marketplace exactly same as our version
(Only whitespaces different)


rikterbeek:
@maassenbas can we remove this file on both releases: metadata/site_import/sites/.DS_Store ?

PR #87, opened 2018-11-06
Secured Fields in checkout

maassenbas:
Displaying Secured Fields in checkout


cartridges/LINK_adyen/cartridge/client/default/js/adyenSecureFields.js
```diff
@@ -0,0 +1,49 @@
+    const configuration = {
+        locale: 'en_US', // Defaults to en_US
```
cyattilakiss: Can you retrieve from the platform the users locale maybe? (It's nicer if you pass the default from the platform)

cartridges/LINK_adyen/cartridge/client/default/js/adyenSecureFields.js
```diff
@@ -0,0 +1,49 @@
...
+
+    function getConfigurationSecureFields() {
+        $.ajax({
+            url: 'Adyen-GetConfigSecuredFields',
+            type: 'get',
+            data: {protocol : window.location.protocol},
+            success: function (data) {
+                setConfigData(data);
+            },
+            complete: function(){
```
cyattilakiss: Without the config data the card component won't even load I think so maybe it would be better to put the functionality from under the complete event handling into the success

cartridges/LINK_adyen/cartridge/controllers/Adyen.js
```diff
@@ -170,6 +171,29 @@ server.get('GetPaymentMethods', server.middleware.https, function (req, res, nex
...
+    var	adyenGetOriginKey = require('*/cartridge/scripts/adyenGetOriginKey');
+    var baseUrl = req.querystring.protocol + "//" + Site.getCurrent().getHttpsHostName();
+    var originKey;
+    var loadingContext = "";
+
+    try {
+        originKey = adyenGetOriginKey.getOriginKey(baseUrl).originKeys;
+        loadingContext = adyenHelper.getLoadingContext();
+    } catch (err) {
+        originKey = [];
```
cyattilakiss: do we log the errors in the system? If we do then we could do it here with the caught exception

lancergr: shouldn't we return a failed response instead of 200 with empty key & context?

cartridges/LINK_adyen/cartridge/templates/default/checkout/billing/creditCardForm.isml
```diff
@@ -1,121 +1,5 @@
+<input type="hidden" class="cardNumber" value="" id="cardNumber" >
```
cyattilakiss: is the class necessary for the hidden input? Also if no value is assigned the value attribute can also be deleted

cartridges/LINK_adyen/cartridge/templates/default/checkout/billing/paymentOptions.isml
```diff
@@ -2,6 +2,8 @@
     var assets = require('*/cartridge/scripts/assets.js');
     assets.addJs('/js/adyen-cse.js');
     assets.addCss('/css/adyenCss.css');
+    assets.addJs('/js/adyenSecureFields.js');
```
cyattilakiss: if you use the card components I think this js is not necessary anymore.
Maybe the same is true for the css under this? (I'm not sure)

metadata/site_import/meta/system-objecttype-extensions.xml
```diff
@@ -360,6 +360,15 @@
                 <min-length>0</min-length>
                 <field-length>0</field-length>
             </attribute-definition>
+             <attribute-definition attribute-id="Adyen_live_prefix">
```
cyattilakiss: do you use the live prefix in this commit already or it is in another ticket?

metadata/site_import/services.xml
```diff
@@ -30,6 +30,11 @@
         <user-id></user-id>
         <password/>
     </service-credential>   
+    <service-credential service-credential-id="AdyenOriginKeys">
+        <url>https://checkout-test.adyen.com/v40/originKeys</url>
```
lancergr: in order to not create a new service-credential for every api call, can't we modify the service factory to append these URLs?
on https://github.com/Adyen/adyen-salesforce-commerce-cloud/blob/36ac2b7cf76586d0dbcda0ae38e59408249c0a14/cartridges/int_adyen/cartridge/scripts/util/AdyenHelper.ds#L41
to do svc.setUrl

Aleffio: let's use v1 for consistency:
```suggestion
        <url>https://checkout-test.adyen.com/v1/originKeys</url>
```

PR #91, opened 2018-11-16
Fixing blocking bugs. Removed fatal null checks and fixed a variable …

sgude:
…name.


maassenbas:
Hi @sgude , 

Thanks for the PR. 
I checked the updated code and see that this is deprecated code that we will delete soon anyway. 
I will close this PR for now. 

Kind regards,
Bas Maassen

PR #103, opened 2018-12-12
Zero auth with Secured Fields

maassenbas:


cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
```diff
@@ -331,15 +338,13 @@ var __AdyenHelper : Object = {
...
-        }
-        if ('additionalData' in cardObject) {
-            jsonObject['additionalData'] = cardObject.additionalData;
-        }
+        var paymentMethodObject = __AdyenHelper.createCardObjectSF(false, 'account');
+        jsonObject['paymentMethod'] = paymentMethodObject;
+
+        jsonObject['recurring'] = {
+            'contract': 'ONECLICK'
+        };
```
rikterbeek: I believe this does not work with checkoutAPI. storeDetails = true should store it as ONECLICK. I think we need to add enableRecurring = true if recurring is enabled. How do we configure ONECLICK,RECURRING at the moment ?

maassenbas: Changed it to enableOneClick=true

PR #115, opened 2019-01-22
Small fixes

maassenbas:
Small fixes after rebase SF and Checkout API


cartridges/int_adyen_SFRA/cartridge/controllers/Adyen.js
```diff
@@ -41,7 +41,7 @@ server.post('AuthorizeWithForm', server.middleware.https, function (req, res, ne
         };
         var result = adyenCheckout.doPaymentDetailsCall(jsonRequest);
         Transaction.wrap(function () {
-            paymentInstrument.custom.adyenPaymentData = "";
+            paymentInstrument.custom.adyenPaymentData = null;
```
ScottPowell14: Is this indeed different?

`paymentInstrument.custom.adyenPaymentData = "";` vs.
`paymentInstrument.custom.adyenPaymentData = null;`

maassenbas: No, it is not different, but nicer to set it to null

PR #116, opened 2019-02-04
Changed session variable

maassenbas:
Session variables changed from Objects to ID's


cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/adyen_credit.js
```diff
@@ -83,8 +86,8 @@ function Authorize(orderNumber, paymentInstrument, paymentProcessor) {
 
     if (result.IssuerUrl != '') {
         Transaction.commit();
-        session.custom.order = order;
-        session.custom.paymentInstrument = paymentInstrument;
+        session.custom.orderNo = order.orderNo;
+        session.custom.paymentMethod = paymentInstrument.paymentMethod;
```
rikterbeek: is paymentMethod unique ?

cartridges/int_adyen_controllers/cartridge/controllers/Adyen.js
```diff
@@ -370,12 +370,20 @@ function cancelOrRefund() {
...
-	var	adyen3DVerification = require('int_adyen_overlay/cartridge/scripts/adyen3DVerification'), result,
-	order = session.custom.order,
-	paymentInstrument = session.custom.paymentInstrument,
+	var	adyen3DVerification = require('int_adyen_overlay/cartridge/scripts/adyen3DVerification'), result, order, paymentInstrument,
 	adyenResponse  = session.custom.adyenResponse;
+
+	if(session.custom.orderNo && session.custom.paymentMethod) {
+		order = OrderMgr.getOrder(session.custom.orderNo);
+		paymentInstrument = order.getPaymentInstruments(session.custom.paymentMethod)[0];
+	}
```
rikterbeek: as this is an array it looks like this is not an unique value ?

cartridges/int_adyen_SFRA/cartridge/controllers/Adyen.js
```diff
@@ -27,58 +27,59 @@ server.get('Adyen3D', server.middleware.https, function (req, res, next) {
...
-      PaResponse: req.form.PaRes
-    });
 
-    // if error, return to checkout page
-    if (result.error || result.Decision != 'ACCEPT') {
-      Transaction.wrap(function () {
-        OrderMgr.failOrder(order);
+  if(session.custom.orderNo && session.custom.paymentMethod) {
+    var order = OrderMgr.getOrder(session.custom.orderNo);
+    var paymentInstrument = order.getPaymentInstruments(session.custom.paymentMethod)[0];
```
cyattilakiss: I would suggest some defensive programming since it can happen that the `order.getPaymentInstruments(session.custom.paymentMethod)` returns undefined and the undefined[0] will return an error

cartridges/int_adyen_SFRA/cartridge/controllers/Adyen.js
```diff
@@ -27,58 +27,59 @@ server.get('Adyen3D', server.middleware.https, function (req, res, next) {
...
+    if (session.custom.MD == req.form.MD) {
```
cyattilakiss: also here check if the paymentInstrument exists

cartridges/int_adyen_controllers/cartridge/controllers/Adyen.js
```diff
@@ -370,12 +370,20 @@ function cancelOrRefund() {
...
 {
-	var	adyen3DVerification = require('int_adyen_overlay/cartridge/scripts/adyen3DVerification'), result,
-	order = session.custom.order,
-	paymentInstrument = session.custom.paymentInstrument,
+	var	adyen3DVerification = require('int_adyen_overlay/cartridge/scripts/adyen3DVerification'), result, order, paymentInstrument,
 	adyenResponse  = session.custom.adyenResponse;
+
+	if(session.custom.orderNo && session.custom.paymentMethod) {
+		order = OrderMgr.getOrder(session.custom.orderNo);
+		paymentInstrument = order.getPaymentInstruments(session.custom.paymentMethod)[0];
```
cyattilakiss: also here - but if you use it more than once maybe you can create a getter function for it where you handle if the instrument does not exits and then you are done everywhere :)

cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/adyen_credit.js
maassenbas: No, it is not unique, but with the paymentMethod, you can retrieve the paymentInstrument of the order

rikterbeek: Can we then not better use a paymentInstrumentId that is unique so we can store this and retrieve it ?

PR #128, opened 2019-02-28
fix metadata for adyencheckoutpaymentmethods

Aleffio:


metadata/site_import/services.xml
```diff
@@ -34,13 +34,18 @@
...
         <url>https://checkout-test.adyen.com/v1/originKeys</url>
         <user-id></user-id>
         <password/>
-    </service-credential>   
-     
+    </service-credential>
+    <service-credential service-credential-id="AdyenCheckoutPaymentMethods">
+        <url>https://checkout-test.adyen.com/v40/paymentMethods</url>
+        <user-id></user-id>
+        <password/>
```
rikterbeek: any possibilities for x-api-key ?

maassenbas: API key is taken from Custom Preferences, so username/password is not required there.

PR #129, opened 2019-03-01
iDeal Component SG

maassenbas:
Handling iDeal component in checkout.
If payment method has issuers, display issuers in selectbox.

### DIFF
--- a/cartridges/int_adyen_overlay/cartridge/js/pages/checkout/adyen-cse.js
+++ b/cartridges/int_adyen_overlay/cartridge/js/pages/checkout/adyen-cse.js
@@ -38,9 +38,7 @@
                 }
                 else if (radioVal == "Adyen"){
                     var selectedMethod = $('[name="brandCode"]:checked').val();
-                    if(!componentDetailsValid(selectedMethod)){
-                        return false;
-                    }
+                    return componentDetailsValid(selectedMethod);
                 }
 
                 e.preventDefault();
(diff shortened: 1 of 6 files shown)


cartridges/int_adyen_overlay/cartridge/js/pages/checkout/adyen-cse.js
```diff
@@ -35,14 +35,19 @@
...
-                    $('.form-data-error').html('');
-                    $('#billing-submit-hidden').trigger('click');
-
                 }
-            });
+                else if (radioVal == "Adyen"){
+                    var selectedMethod = $('[name="brandCode"]:checked').val();
+                    if(!componentDetailsValid(selectedMethod)){
+                        return false;
+                    }
```
rikterbeek: don't think this if statement is needed. I believe componentDetailsValid will provide you back true/false. So you can simplify by return componentDetailsValid(selectedMethod)

cartridges/int_adyen_overlay/cartridge/js/pages/checkout/adyen-cse.js
```diff
@@ -19,7 +19,7 @@
             var isOneClick = false;
             $('#billing-submit').on('click', function (e) {
                 var radioVal = $('.payment-method-options').find(':checked').val();
-                if ('CREDIT_CARD' == radioVal) {
+                if (radioVal == 'CREDIT_CARD') {
```
ScottPowell14: lol why?

cartridges/int_adyen_overlay/cartridge/js/pages/checkout/adyen-cse.js
maassenbas: You are right, will update it

PR #133, opened 2019-03-11
SFRA Open Invoice component

maassenbas:
Checkout component for SFRA


cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -165,9 +169,101 @@
...
+     * @returns {*}
+     */
+    function filterUndefinedItemsInArray(arr) {
+        return arr.filter(function (item) {
+            return typeof item !== 'undefined';
+        });
+    };
+
+    function getSsnLengthNordicCountry(country){
+        if(country.indexOf("NO") > -1 || country.indexOf("SE") > -1 || country.indexOf("FI") > -1 || country.indexOf("DK") > -1){
```
cyattilakiss: Do we need to check it with the indexof? I mean we can simply check if 

```suggestion
        if(country === "NO" || country === "SE"|| country === "FI" || country === "DK"){
```

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -165,9 +169,101 @@
...
+            if(country.indexOf("NO") > -1){
```
cyattilakiss:
```suggestion
            if(country === "NO"){
```

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -165,9 +169,101 @@
...
+     */
+    function filterUndefinedItemsInArray(arr) {
+        return arr.filter(function (item) {
+            return typeof item !== 'undefined';
+        });
+    };
+
+    function getSsnLengthNordicCountry(country){
+        if(country === "NO" || country === "SE"|| country === "FI" || country === "DK"){
+            if(country === "NO"){
```
rkewlani: We can remove additional check for country ==NO. it is being evaluated twice. Something like :
if(country==NO) return 5;
if (country ==SE || country ==FI || country==DK) return 4;
else return 0

PR #136, opened 2019-03-20
Hide BCMC mobile and WeChat QR

maassenbas:


cartridges/int_adyen_SFRA/cartridge/controllers/Adyen.js
```diff
@@ -167,6 +167,18 @@ server.get('GetPaymentMethods', server.middleware.https, function (req, res, nex
   return next();
 });
 
+/**
+ * Checks if payment method is blocked
+ */
+function isMethodTypeBlocked(methodType)
```
rkewlani: Not sure if it will return false or something else in case condition is not meet. Can you please add else {return false} if it does not return false by default.

PR #143, opened 2019-04-09
Added support for Oney as Open Invoice

maassenbas:


cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
```diff
@@ -241,7 +241,7 @@ var __AdyenHelper : Object = {
     },
 
     isOpenInvoiceMethod: function(paymentMethod){
-            if(paymentMethod.indexOf("afterpay") > -1 || paymentMethod.indexOf("klarna") > -1 || paymentMethod.indexOf("ratepay") > -1 ){
+            if(paymentMethod.indexOf("afterpay") > -1 || paymentMethod.indexOf("klarna") > -1 || paymentMethod.indexOf("ratepay") > -1 || paymentMethod.indexOf("facilypay") > -1 ){
```
KadoBOT: You can use the `includes` method as well. I think it works best in this case. So you would have something like:
`if(paymentMethod.includes("afterpay") ||...`

PR #149, opened 2019-05-09
Compatibility with SFRA v4.0.0

maassenbas:


rikterbeek:
@maassenbas did we now remove all complete overwrites of core code ?

maassenbas:
No, we removed the append from SubmitPayment, but will still replace the PlaceOrder action

PR #173, opened 2019-07-03
Refactor code for My Account

maassenbas:
Created reusable code for billing/account checkout

### DIFF
--- a/cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
+++ b/cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
@@ -390,39 +390,25 @@ var __AdyenHelper: Object = {
 
     createAddressObjects: function (order, jsonObject) {
         var shippingAddress = order.defaultShipment.shippingAddress;
-        var houseNumberShip;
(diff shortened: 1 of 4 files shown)


cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
```diff
@@ -313,67 +313,122 @@ var __AdyenHelper: Object = {
...
+        if (shippingAddress.address2) {
+            houseNumberShip = shippingAddress.address2;
         }
+        else {
+            houseNumberShip = "N/A";
+        }
+        jsonObject["countryCode"] = shippingAddress.countryCode.value.toUpperCase();
 
-        return shopperObject;
+        jsonObject.deliveryAddress = {
```
rikterbeek: Improvement here can be that we set the defaults first and then override with the actual value see: https://github.com/Adyen/adyen-magento2/blob/develop/Helper/Requests.php#L157-L163

maassenbas: Done

PR #178, opened 2019-07-10
Add detailed stamping to avoid collisions

jkol:
Notifications received in rapid succession occasionally collide and are dropped from being created due to more than one notification being received in the same second. Adding milliseconds prevents these collisions.


jkol:
Was getting this error log:

> Error message: com.demandware.beehive.core.capi.domain.meta.MetaDataException: Couldn't create custom object instance of type 'adyenNotification' key '00006018-20190709095846. Key is not unique.

maassenbas:
Hi @jkol , 
Thanks for the fix!
I will merge the PR now.

PR #182, opened 2019-07-18
Filter expresspay and applepay, cup on component

maassenbas:


cartridges/int_adyen_SFRA/cartridge/controllers/Adyen.js
```diff
@@ -257,23 +264,22 @@ server.get('GetPaymentMethods', server.middleware.https, function (req, res, nex
         countryCode = currentBasket.getShipments()[0].shippingAddress.getCountryCode();
     }
     var paymentMethods;
+    var descriptions = [];
     try {
         paymentMethods = getPaymentMethods.getMethods(BasketMgr.getCurrentBasket(), countryCode.value.toString()).paymentMethods;
+        paymentMethods = paymentMethods.filter(function (method) {
+            return !isMethodTypeBlocked(method.type);
+        });
+        paymentMethods.forEach(function (method) {
```
KadoBOT: Instead of pushing to descriptions, you could have something like:
```suggestion
        var descriptions = paymentMethods.map(function (method) {
```

PR #190, opened 2019-08-08
Voucher component

maassenbas:
Konbini, Dragonpay


cartridges/int_adyen_SFRA/cartridge/templates/default/checkout/billing/paymentOptions/alternativePaymentMethodSummary.isml
```diff
@@ -1,3 +1,23 @@
...
+    <div id="voucherResult"></div>
+    <br/>
+    <script type="text/javascript">
+        var configuration = {
+            locale: '${request.locale}'
+        };
+
+        const checkout = new AdyenCheckout(configuration);
+        const voucherResult = document.getElementById('voucherResult');
+        var action = JSON.parse('${payment.adyenAction}'.replace(/&quot;/g, '\"'));
```
rikterbeek: looks a bit nasty with .replace(/&quot;/g, '\"') is there no better solution possible ? @KadoBOT can you have a look as well ?

cartridges/int_adyen_overlay/cartridge/templates/resources/checkout.properties
```diff
@@ -8,4 +8,7 @@ comprafacil.description=Please pay with the provided Multibanco reference and en
 comprafacil.deadline=Payment Deadline (in days)
 comprafacil.reference=Reference
 comprafacil.entity=Entity
-comprafacil.amount=Amount
\ No newline at end of file
+comprafacil.amount=Amount
+econtext.payCode=支払の参照
+econtext.maskedTelephoneNumber=電話番号
+econtext.payLimitDateTime=有効期限
```
rikterbeek: should this not be in english and all countries have there own language file ?

cartridges/int_adyen_SFRA/cartridge/templates/default/checkout/billing/paymentOptions/alternativePaymentMethodSummary.isml
KadoBOT: I don't see any benefit of using the &quot; in first place nowadays. But if someone is using it, and we need to replace it with `"` this is the way to do it.

KadoBOT: I just don't get why `payment.adyenAction` is wrapped in `'${}'`. What about the code below:
`var action = JSON.parse(payment.adyenAction.replace(/&quot;/g, '\"'));`

cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.ds
```diff
@@ -232,6 +232,24 @@ function doPaymentCall(args, requestObject){
                 Logger.getLogger('Adyen').debug('Adyen: ' + resultObj.statusCode + ' Error => ' + resultObj.statusMessage + ' | ' + resultObj.errorText);
             }
         }
+        else if(responseObj.resultCode.indexOf('PresentToShopper') != -1) {
```
cyattilakiss: I found it several times that we are checking if the string starts with an expected string but why don't we just check if the string is the expected string or not?
```suggestion
        else if(responseObj.resultCode == 'PresentToShopper') {
```

maassenbas: Good one, I will update it

cartridges/int_adyen_SFRA/cartridge/templates/default/checkout/billing/paymentOptions/alternativePaymentMethodSummary.isml
maassenbas: No other solution

PR #207, opened 2019-10-04
Order.custom.Adyen_value has decimal when set.

bhunter:
Added toFixed(0) so order's field value is a whole number when parsed to string.

Looks like you have wrestled with this in the past.
Is confusing in Business Manager and can cause (downstream) notification processing errors because field will get parsed with decimal in it then amount will be wonky.

### DIFF
--- a/cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
+++ b/cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
@@ -664,7 +664,7 @@ var __AdyenHelper: Object = {
         }
 
         if ('AdyenAmount' in result && !empty(result.AdyenAmount)) {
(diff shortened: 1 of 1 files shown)


maassenbas:
Hi @bhunter , 
Do you have any examples where it went wrong? 
I'm trying to identify why this .toFixed(0) is needed, but I don't see the added value. 
Could you explain it a bit more? 
Thanks,
Bas

bhunter:
Hey Bas,

While it is not a crippling bug, the reporting in Business Manager is incorrect, the value is supposed to represent minor units and with a decimal it does not. Plus, if other Adyen endpoints are called -such as /capture (from a notification), you cannot use the amount in the field without some correcting calculations. Here is a screenshot of my admin when the .toFixed(0) is not present.

Screen Shot 2019-10-08 at 9 09 29 AM

PR #208, opened 2019-10-07
Added response redirectResult

maassenbas:


cartridges/int_adyen_controllers/cartridge/controllers/Adyen.js
```diff
@@ -63,26 +63,38 @@ function redirect(order, redirectUrl) {
...
         adyenPaymentInstrument = instrumentsIter.next();
-        Transaction.wrap(function () {
-            adyenPaymentInstrument.custom.adyenPaymentData = null;
-        });
+        paymentData = adyenPaymentInstrument.custom.adyenPaymentData;
     }
 
+    //details is either redirectResult or payload
+    var details;
+    if(request.httpParameterMap.redirectResult.value != null){
```
rikterbeek: will this work if redirectResult is undefined or emtpy because you do a .value to this ?

maassenbas: Yes, then it will go into the else if

PR #213, opened 2019-10-14
Get correct OriginKey with aliases

maassenbas:


cartridges/int_adyen_SFRA/cartridge/templates/default/account/payment/paymentForm.isml
```diff
@@ -17,8 +17,15 @@
             </isscript>
             <iscontent type="text/html" encoding="off"/>
             <script src="${pdict.AdyenHelper.getCheckoutUrl()}" type="text/javascript"></script>
+            <isset name="adyenOriginKey" value="${pdict.AdyenHelper.getOriginKey()}" scope="page"></isset>
```
rikterbeek: should the pdict not just contain the originKey itself ? Now you are calling a helper that implements a method that does the API call. I would suggest to do this in the backend and set it into the pdict value so you only have to read the variable rather then to execute the logic.

maassenbas: Yes, that is a better approach. I updated it.

PR #221, opened 2019-12-17
Show validations for card component

maassenbas:


cartridges/int_adyen_SFRA/cartridge/client/default/scss/adyenCss.scss
```diff
@@ -30,6 +30,10 @@
   display: none;
 }
 
+.adyen-checkout__dropdown__list {
+  z-index: 10 !important;
```
msilvagarcia: Using `!important` clauses are a really bad practice. Better overwrite with with a more specialized selector.

PR #229, opened 2020-01-10
Added BCMC to blocklist for APMs

maassenbas:
BCMC should be processed through the credit card component and not listed in the Alternative Payment Methods (APM) list. Hence, we will prevent it from displaying there.

### DIFF
--- a/cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
+++ b/cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
@@ -610,6 +610,19 @@ var __AdyenHelper: Object = {
         return jsonObject;
     },
 
+    /**
+     * Checks if payment method is blocked
+     */
+    isMethodTypeBlocked : function(methodType) {
+        if (methodType.indexOf('bcmc') !== -1 ||
+            (methodType.indexOf('wechatpay') !== -1 && methodType.indexOf('wechatpayWeb') === -1) ||
+            methodType == "scheme" || methodType == "cup" || methodType == "applepay") {
+            return true;
+        } else {
+            return false;
+        }
+    },
+
     extend: function (obj, src) {
         for (var key in src) {
             if (src.hasOwnProperty(key)) obj[key] = src[key];
--- a/cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
+++ b/cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
@@ -614,8 +614,8 @@ var __AdyenHelper: Object = {
      * Checks if payment method is blocked
      */
     isMethodTypeBlocked : function(methodType) {
-        if (methodType.indexOf('bcmc') !== -1 ||
-            (methodType.indexOf('wechatpay') !== -1 && methodType.indexOf('wechatpayWeb') === -1) ||
+        if (methodType == "bcmc" || methodType == "bcmc_mobile_QR" ||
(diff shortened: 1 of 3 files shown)


cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
```diff
@@ -610,6 +610,19 @@ var __AdyenHelper: Object = {
         return jsonObject;
     },
 
+    /**
+     * Checks if payment method is blocked
+     */
+    isMethodTypeBlocked : function(methodType) {
+        if (methodType.indexOf('bcmc') !== -1 ||
+            (methodType.indexOf('wechatpay') !== -1 && methodType.indexOf('wechatpayWeb') === -1) ||
+            methodType == "scheme" || methodType == "cup" || methodType == "applepay") {
```
rikterbeek: are we not missing bcmc_mobile_QR ?

maassenbas: This will also be included with "methodType.indexOf('bcmc')"

cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
```diff
@@ -614,8 +614,8 @@ var __AdyenHelper: Object = {
      * Checks if payment method is blocked
      */
     isMethodTypeBlocked : function(methodType) {
-        if (methodType.indexOf('bcmc') !== -1 ||
-            (methodType.indexOf('wechatpay') !== -1 && methodType.indexOf('wechatpayWeb') === -1) ||
+        if (methodType == "bcmc" || methodType == "bcmc_mobile_QR" ||
```
KadoBOT: What about? `if (["bcmc", "bcmc_mobile_QR"].indexOf(methodType) !== 1 ||`

cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
```diff
@@ -614,8 +614,8 @@ var __AdyenHelper: Object = {
...
+            (methodType.indexOf("wechatpay") !== -1 && methodType.indexOf("wechatpayWeb") === -1) ||
```
KadoBOT: Isn't this the same as `methodType === 'wechatpay' ?

cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
cyattilakiss: Also you could create an array before the if with all the unsupported payment methods and just check as KadoBOT suggested if it is in the array or not. That is a bit more readable solution I think

cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
cyattilakiss: no because there are some other wechatpay payment methods that we want to include in the unsupported payment methods list. In prestashop we listed all the method names and didn't just check the first part of the payment methods. Personally I like this approach more: https://github.com/Adyen/adyen-prestashop/blob/1.2.0/adyen.php#L1036 Easier to understand and gives a better overview of the payment methods.

KadoBOT: Ah gotcha, there are 4 `wechatpay` 👌

PR #231, opened 2020-01-20
Add /connectedTerminals

maassenbas:
Get connected terminals and display them in a dropdown in the checkout page


cartridges/int_adyen_overlay/cartridge/scripts/adyenGetConnectedTerminals.ds
```diff
@@ -0,0 +1,74 @@
...
+
+        var xapikey = AdyenHelper.getAdyenApiKey();
+        service.addHeader("Content-type", "application/json");
+        service.addHeader("charset", "UTF-8");
+        service.addHeader("X-API-key", xapikey);
+        callResult = service.call(JSON.stringify(requestObject));
+
+        if (callResult.isOk() == false) {
+            Logger.getLogger("Adyen").error("Adyen: Call error code" +  callResult.getError().toString() + " Error => ResponseStatus: " + callResult.getStatus()  + " | ResponseErrorText: " +  callResult.getErrorMessage() + " | ResponseText: " + callResult.getMsg());
+            return ;
```
msilvagarcia: This could throw an error instead, so that you would have a centralized logging code.

cartridges/int_adyen_overlay/cartridge/scripts/adyenGetConnectedTerminals.ds
```diff
@@ -0,0 +1,74 @@
...
+    try {
+        var requestObject = {};
+
+        var callResult = null,
+            resultObject = null,
+            service = AdyenHelper.getService(AdyenHelper.SERVICE.CONNECTEDTERMINALS);
+
+        if (service == null) {
+            Logger.getLogger("Adyen").error("Could not retrieve connected terminals, error creating the service");
+            return;
```
msilvagarcia: This could throw an error instead, so that you would have a centralized logging code.

cartridges/int_adyen_overlay/cartridge/scripts/adyenGetConnectedTerminals.ds
```diff
@@ -0,0 +1,74 @@
...
+            return ;
+        }
+
+        try {
+            return JSON.parse(resultText);
+        }
+        catch (ex) {
+            Logger.getLogger("Adyen").error("Adyen: error parsing response object " + ex.message);
+            return;
+        }
```
msilvagarcia: You can move this outside of the try-catch block so you don't have them nested.

cartridges/int_adyen_overlay/cartridge/scripts/adyenGetConnectedTerminals.ds
KadoBOT: Also, what if `service` is `undefined` or `false`? What about `if(!service)` ?
Same applies for the other ifs that compares to `null`

cartridges/int_adyen_overlay/cartridge/scripts/adyenGetConnectedTerminals.ds
```diff
@@ -0,0 +1,74 @@
...
+        service.addHeader("charset", "UTF-8");
+        service.addHeader("X-API-key", xapikey);
+        callResult = service.call(JSON.stringify(requestObject));
+
+        if (callResult.isOk() == false) {
+            Logger.getLogger("Adyen").error("Adyen: Call error code" +  callResult.getError().toString() + " Error => ResponseStatus: " + callResult.getStatus()  + " | ResponseErrorText: " +  callResult.getErrorMessage() + " | ResponseText: " + callResult.getMsg());
+            return ;
+        }
+
+        resultObject = ("object" in callResult ? callResult.object : null);
```
KadoBOT: Whats the difference between this and `resultObject = callResult.object` ? If `object` is not available, `resultObject` will be `undefined`... which is fine if you don't compare stuff with `null`.

cartridges/int_adyen_overlay/cartridge/scripts/adyenGetConnectedTerminals.ds
```diff
@@ -0,0 +1,74 @@
...
+
+        var resultObj = {
+            statusCode: resultObject.getStatusCode(),
+            statusMessage: resultObject.getStatusMessage(),
+            text: resultObject.getText(),
+            errorText: resultObject.getErrorText(),
+            timeout: resultObject.getTimeout()
+        }
+
+        var resultText = ("text" in resultObj && !empty(resultObj.text) ? resultObj.text : null);
```
KadoBOT: Same as previous comment.

PR #243, opened 2020-03-06
PW-1540 Setting up Babel, Unit testing environment, and github actions

zaiddreakh:
Setting up Babel, Unit testing environment, and github actions


setup-jest.js
```diff
@@ -0,0 +1,14 @@
+//importing jquery
```
KadoBOT: I feel that this comment (and the ones below) are unnecessary. The code explains itself, and those comments just adds noise

src/cartridges/int_adyen_SFRA/test/unit/scripts/checkout/adyenCheckout.test.js
```diff
@@ -0,0 +1,8 @@
+const adyenCheckout = require('../../../../cartridge/client/default/js/adyenCheckout');
```
KadoBOT: should this be here?

PR #248, opened 2020-03-24
PW-2152 Boleto Generic Component

maassenbas:


cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -1,12 +1,51 @@
...
+            }
+        },
+        boletobancario: {
+            personalDetailsRequired: true, // turn personalDetails section on/off
+            billingAddressRequired: false, // turn billingAddress section on/off
+            showEmailAddress: false, // allow shopper to specify their email address
+
+            // Optionally prefill some fields, here all fields are filled:
+            data: {
+                socialSecurityNumber: '56861752509',
```
KadoBOT: Just make sure to remove this mocked data when merging to develop. Maybe add a // TODO here?

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -1,12 +1,51 @@
...
+                firstName: 'José',
+                lastName: 'Silva',
+            }
+        },
+    };
+const checkout = new AdyenCheckout(checkoutConfiguration);
```
KadoBOT: Shouldn't this be a `var` ? I see `const` in other places as well. Just make sure to use one or another, not both. And if `const`/`let` is already supported, I suggest getting rid of the `var`s

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -51,27 +90,14 @@ function renderGenericComponent() {
             } else {
                  try {
                      var node = checkout.create(paymentMethod.type, {
-                         details: paymentMethod.details,
-                         enableStoreDetails: showStoreDetails,
                          onChange: function (state) {
```
KadoBOT: Shouldn't this be in the `paymentMethodsConfiguration` object?

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -1,12 +1,51 @@
...
+        boletobancario: {
+            personalDetailsRequired: true, // turn personalDetails section on/off
+            billingAddressRequired: false, // turn billingAddress section on/off
+            showEmailAddress: false, // allow shopper to specify their email address
+
+            // Optionally prefill some fields, here all fields are filled:
+            data: {
+                socialSecurityNumber: '56861752509',
+                firstName: 'José',
+                lastName: 'Silva',
```
rikterbeek: can we prefill this data with the actual firstname lastname filled into the checkout ?

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
rikterbeek: as well the onChange event you handle on the highest level we should combine that. ideally we have everything on root level as much as possible.

KadoBOT: Also, the `paymentMethod.type` is dynamic I assume? But the logic inside the options applies only to type `card`. This might cause some bugs.
Moving to the `card` object inside the root config should solve this issue.

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
cyattilakiss: Checkout component still supports browsers where const/let is not supported, that was the reason I had to revert all these changes on M2 regarding these :( But if SFCC doesn't anyway then of course you could and should ;)

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
maassenbas: Good idea, will do

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
maassenbas: I will update this to var

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -2,43 +2,47 @@ var checkoutConfiguration = window.Configuration;
...
+        }
+    },
+    boletobancario: {
+        personalDetailsRequired: true, // turn personalDetails section on/off
+        billingAddressRequired: false, // turn billingAddress section on/off
+        showEmailAddress: false, // allow shopper to specify their email address
+
+        // Optionally prefill some fields, here all fields are filled:
+        data: {
+            socialSecurityNumber: '56861752509',
```
KadoBOT: this key should be removed

maassenbas: Yes you are right. Will remove it once we merge it to develop

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
maassenbas: Moved everything now to the configuration object in the root

PR #250, opened 2020-03-27
Pw-535 Credit Card Installments

zaiddreakh:
Add credit card installments for SFRA based on Generic Component


cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -1,6 +1,19 @@
...
 var isValid = false;
 var storeDetails;
+var cardNode = document.getElementById('card');
+var oneClickCard = [];
+var card;
+var idealComponent;
+var afterpayComponent;
+var klarnaComponent;
+var maskedCardNumber;
+const MASKED_CC_PREFIX = '************';
```
KadoBOT: `var` ?

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -43,18 +56,8 @@ checkoutConfiguration.paymentMethodsConfiguration = {
...
-var card;
-var idealComponent;
-var afterpayComponent;
-var klarnaComponent;
-var maskedCardNumber;
-const MASKED_CC_PREFIX = '************';
-var oneClickValid = false;
-var selectedMethod;
-var componentArr = [];
+if(installments)
```
KadoBOT:
```suggestion
if(window.installments) {
var installments = JSON.parse(window.installments)
```
Since `installments` is a string you don't need line 16. This should be enough.

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -68,6 +71,10 @@ function displaySelectedMethod(type) {
 function renderGenericComponent() {
     getPaymentMethods( function (data) {
         checkoutConfiguration.paymentMethodsResponse = data.AdyenPaymentMethods;
+        if(data.amount)
```
KadoBOT: Try to always use `{` on `if` blocks, just for good practice.

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -1,6 +1,19 @@
...
+var afterpayComponent;
+var klarnaComponent;
+var maskedCardNumber;
+const MASKED_CC_PREFIX = '************';
+var oneClickValid = false;
+var selectedMethod;
+var componentArr = [];
+
+var checkoutConfiguration = window.Configuration;
+var installments = window.installments.length > 0 ? JSON.parse(window.installments) : null;
```
cyattilakiss: I think it's better to use if statement instead of the conditional operator. Easier to read

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
KadoBOT: Also, wrap `JSON.parse` in a `try...catch`

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
KadoBOT: > I think it's better to use if statement instead of the conditional operator. Easier to read

I have to disagree there

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -95,7 +102,9 @@ function renderGenericComponent() {
                      node.mount(container);
                      componentArr[paymentMethod.type] = node;
                  }
-                 catch (e) {}
+                 catch (e) {
+                    console.error(e);
```
maassenbas: Do we have to log this when there is an error? Maybe we can just skip it and proceed with next component instead.

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -298,7 +307,8 @@ function getFallback(paymentMethod) {
                     <span class="adyen-checkout__label">Date of birth</span>
                     <input id="dateOfBirthInput" class="adyen-checkout__input" type="date"/>`;
 
-    var fallback = {ach: ach, ratepay: ratepay};
+    var paysafecard = ' ';
+    var fallback = {ach: ach, ratepay: ratepay, paysafecard: paysafecard};
```
maassenbas: Regarding paysafecard, if there is no content in the fallback, I would suggest not to include it here.

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -56,8 +55,10 @@ checkoutConfiguration.paymentMethodsConfiguration = {
         }
     }
 };
-if(installments)
+if(window.installments) {
+    var installments = JSON.parse(window.installments);
```
KadoBOT: just add a `try...catch` around the JSON.parse and your good to merge :)

PR #253, opened 2020-04-08
PW-1995 /paymentmethods call for OneClick payments

zaiddreakh:
/paymentmethods call for OneClick payments in SFRA


cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -33,13 +29,17 @@ checkoutConfiguration.paymentMethodsConfiguration = {
...
         },
-        onChange: function (state) {
+        onChange: function (state, component) {
             storeDetails = state.data.storePaymentMethod;
             isValid = state.isValid;
             var type = state.data.paymentMethod.type;
-            $('#browserInfo').val(JSON.stringify(state.data.browserInfo));
-            componentArr[type].isValid = isValid;
-            componentArr[type].stateData = state.data;
+            var componentName = component._node.id.replace('component_', '');
```
KadoBOT, line 36: Because of the `_` in front of the object, I would assume that this is not meant to be publicly used. Which might cause bugs in the future, so be aware of that. Maybe add a comment in the code?

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -84,6 +83,41 @@ function renderGenericComponent() {
         var checkout = new AdyenCheckout(checkoutConfiguration);
         document.querySelector("#paymentMethodsList").innerHTML = "";
         var paymentMethodsUI = document.querySelector('#paymentMethodsList');
+
+        if(data.AdyenPaymentMethods.storedPaymentMethods) {
+            for (var i = 0; i < data.AdyenPaymentMethods.storedPaymentMethods.length; i++) {
+                var paymentMethod = data.AdyenPaymentMethods.storedPaymentMethods[i];
```
maassenbas, line 89: We should add a filter on supportedShopperInteractions = Ecommerce in this loop. This will indicate the OneClick contracts. I assume that this now also renders Recurring contracts (supportedShopperInteractions = ContAuth), which we don't want to display on the checkout page

cartridges/int_adyen_SFRA/cartridge/client/default/js/checkout/checkout.js
```diff
@@ -17,7 +17,9 @@ module.exports = {
                     data.options
                 );
             });
-            adyenCheckout.methods.renderGenericComponent();
+            var currentStage = location.search.substring(location.search.indexOf("=")+1);
+            if(currentStage === "shipping")
```
maassenbas, line 21: We have to keep in mind that virtual products don't have a shipping stage, but we do have to render the payment methods. We must think of a solution for that.

cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -63,8 +63,7 @@ if(window.installments) {
     catch (e) {}
 }
 
-// renderOneClickComponents();
-renderGenericComponent();
+// renderGenericComponent();
```
maassenbas, line 66: We can delete this line if we don't call the renderGenericComponent() from here

PR #255, opened 2020-04-09
Calculate item amount for price adjustment using netPrice

xapu88:
While working on integration of ZipPay payments via Adyen, I noticed that the redirection to Zip doesn't work if there is a discount line item in our cart (discount coupon applied). After debugging I came to conclusion that the amounts of the line items which are "instanceof dw.order.PriceAdjustment" aren't calculated well. The "amountExcludingTax" was actually the full amount, because it was retrieved from the basePrice. After using netPrice for the "amountExcludingTax", the calculations are fine.

Note: the prices on the market are inclusive of taxes (GST Inclusive).

### DIFF
--- a/src/cartridges/int_adyen_overlay/cartridge/scripts/util/LineItemHelper.ds
+++ b/src/cartridges/int_adyen_overlay/cartridge/scripts/util/LineItemHelper.ds
@@ -61,7 +61,7 @@ var __LineItemHelper  = {
         if (lineItem instanceof dw.order.ProductLineItem || lineItem instanceof dw.order.ShippingLineItem) {
             return AdyenHelper.getCurrencyValueForApi(lineItem.adjustedNetPrice);
         } else if (lineItem instanceof dw.order.PriceAdjustment) {
-            return AdyenHelper.getCurrencyValueForApi(lineItem.basePrice);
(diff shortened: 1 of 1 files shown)


maassenbas:
Hi @xapu88 , 

Thanks for you PR. 
I am looking into this and will merge the PR once I verified it.
I will keep you updated.

Kind regards,
Bas Maassen
Adyen

maassenbas:
Hi @xapu88 , 

I tried to reproduce your issue, but I am not able to.
For both the basePrice and netPrice, the redirect to Zip pay is working fine for me. 
Do you have a specific example which we can look into? Perhaps a PSP reference so we can look into our log files as well.

Kind regards,
Bas Maassen

xapu88:
Hi Bas,

The PSP reference is: 851586442271221H
In the attachments I will provide you with full request and response.

and the response to /payments/details call was:

{
"pspReference": "851586442271221H",
"resultCode": "Refused",
"merchantReference": "00013405",
"paymentMethod": "zip",
"shopperLocale": "en_AU"
}

Regards

maassenbas:
Thanks @xapu88 . 
I tested your suggestion and verified it is working correctly. 
I will merge this PR. 

Kind regards,
Bas Maassen

PR #262, opened 2020-05-18
Validate state.data fields sent by the frontend

KadoBOT:
Filter out invalid fields from `state.data` object

### DIFF
--- a/cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
+++ b/cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
@@ -436,6 +436,8 @@ var __AdyenHelper: Object = {
 
     createAdyenRequestObject: function (order, paymentInstrument) {
         var jsonObject = JSON.parse(paymentInstrument.custom.adyenPaymentData);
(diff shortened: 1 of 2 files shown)


cartridges/int_adyen_overlay/cartridge/scripts/util/AdyenHelper.ds
```diff
@@ -664,7 +666,27 @@ var __AdyenHelper: Object = {
...
+        for (var i = 0; i < stateDataKeys.length; i++) {
+            var keyName = stateDataKeys[i]
+            var includesInvalidField = validFields.indexOf(keyName) === -1
+            if (includesInvalidField) {
+                invalidFields.push(keyName)
+            } else {
+                filteredStateData[keyName] = stateData[keyName]
+            }
+        }
+        return { stateData: filteredStateData, invalidFields: invalidFields }
```
cyattilakiss, line 687: do you use the 'invalidFields' for any purpose? Maybe it is worth to log it for debugging purposes.

KadoBOT: Currently, it's not used anywhere, but I thought it would be nice to have them saved somewhere if we need in the future.

PR #295, opened 2020-06-29
Updated ds to js

maassenbas:
Renamed all files from *.ds to *.js as suggested during the SFCC demo
## Tested scenarios

**Fixed issue**:


src/cartridges/int_adyen_SFRA/cartridge/scripts/adyenService.js
```diff
@@ -0,0 +1,40 @@
...
+  confirmationEmail.setContent(content, "text/html", "UTF-8");
+  confirmationEmail.send();
+
+  Transaction.wrap(function () {
+    order.custom.Adyen_CustomerEmail = null;
+  });
+}
+
+module.exports = {
+  submit: submit,
```
zaiddreakh, line 39: after the release, we can test the es6 "submit" only without colons

PR #320, opened 2020-07-28
unit tests for Show Confirmation

KadoBOT:
Unit test for `showConfirmationPaymentFromComponent.js` middleware


src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/showConfirmationPaymentFromComponent.js
```diff
@@ -0,0 +1,118 @@
...
+  try {
+    const stateData = JSON.parse(req.form.additionalDetailsHidden);
+    const order = OrderMgr.getOrder(session.privacy.orderNo);
+    const paymentInstruments = order.getPaymentInstruments(
+      constants.METHOD_ADYEN_COMPONENT,
+    );
+    let adyenPaymentInstrument;
+
+    const { paymentData } = stateData;
+    const { details } = stateData;
```
maassenbas, line 24: Why do we assign both variables to stateData?

KadoBOT: This indeed can be reduced, but I've just moved the code. Can be improved during the refactoring.

PR #325, opened 2020-07-29
Sanitize and signature IssuerUrl

maassenbas:
## Tested scenarios

**Fixed issue**:


src/cartridges/int_adyen_SFRA/cartridge/controllers/Adyen.js
```diff
@@ -27,14 +27,25 @@ server.get(
...
+        MD.substr(1, 25),
+    );
+
+    if(signature === currentSignature) {
+      res.render('adyenform', {
+        issuerUrl: IssuerURL,
+        paRequest: PaRequest,
+        md: MD,
+        ContinueURL: TermURL,
+      });
```
rikterbeek, line 43: An easier and better solution would be to retrieve the issuerUrl, paRequest, md, not from the querystring but from the session.

zaiddreakh: We thought about using the session but we feel it's not the most reliable. Some merchants are reporting issues that their shoppers are losing their session and their baskets are emptied, especially due to browser issues. So eventually we decided to keep the querystrings and with the signature they should be secure

src/cartridges/int_adyen_overlay/cartridge/templates/default/adyenform.isml
```diff
@@ -1,24 +1,23 @@
...
 				document.getElementById('3dform').submit();
-				loaded = true;
 			}
 			load();
+
+			function sanitizeHTML(text) {
+				var element = document.createElement('div');
+				element.innerText = text;
+				return element.innerHTML;
+			}
```
rikterbeek, line 20: can't we use jquery for this ?

zaiddreakh: We're trying to reduce jQuery usage in our codebase so eventually we don't use it at all.
the outcome would be the same with jQuery right?

maassenbas: We try to avoid using jQuery where possible. We have now most in vanilla JavaScript.

src/cartridges/int_adyen_SFRA/cartridge/controllers/Adyen.js
rikterbeek: Can we investigate where else we use querystring because everywhere this is a vulnerability if we do not wrap it with a signature we have an issue.

maassenbas:
Hi @sdanilchenko ,

There are no changes required for SiteGenesis, this is only applicable for SFRA.

Kind regards,
Bas Maassen
Adyen

dmersiowsky:
Hi Bas, does that mean the security problem does not exist for the SG integration?

maassenbas:
@dmersiowsky Yes, that is correct. 
The security issue was only in SFRA and not in SiteGenesis.

dmersiowsky:
Great, thanks for confirming!

sdanilchenko:
Hi @maassenbas ,
Thank you!

PR #328, opened 2020-07-30
unit tests for confirm

KadoBOT:
Unit tests for Order `confirm` middleware


src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/confirm.js
```diff
@@ -0,0 +1,49 @@
...
+    const donationAmounts = {
+      currency: session.currency.currencyCode,
+      values: configuredAmounts,
+    };
+
+    const viewData = res.getViewData();
+    viewData.adyen = {
+      originKey,
+      environment,
+      adyenGivingAvailable: true,
```
maassenbas, line 35: We might want to look into this to check whether it is available and not set it always to true

PR #329, opened 2020-07-31
Feat/payment instruments test

KadoBOT:
unit tests for `PaymentInstruments` middlewares


src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/deletePayment.js
```diff
@@ -0,0 +1,25 @@
+const CustomerMgr = require('dw/customer/CustomerMgr');
+const AdyenHelper = require('*/cartridge/scripts/util/adyenHelper');
```
maassenbas, line 2: Should we use adyenHelper instead of AdyenHelper?

KadoBOT: We can do that while refactoring. I've tried to keep everything intact.

PR #331, opened 2020-08-02
PW-2701 Part1: Unit tests for adyen_component

zaiddreakh:
## Tested scenarios

**Fixed issue**:


src/cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/middlewares/__tests__/authorize.test.js
```diff
@@ -0,0 +1,117 @@
+/* eslint-disable global-require */
+let authorize;
+let currentBasket;
+
+beforeEach(() => {
+  authorize = require('../authorize');
+  jest.clearAllMocks();
+  const { getCurrentBasket } = require('dw/order/BasketMgr');
+  currentBasket = getCurrentBasket();
+});
```
KadoBOT, line 10: Here you are using two lines, so it doesn't make sense to use destructuring. You could do this instead:
```js
let currentBasket;

beforeEach(() => {
  authorize = require('../authorize');
  jest.clearAllMocks();
  currentBasket  = require('dw/order/BasketMgr').getCurrentBasket();
});
```

src/cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/middlewares/__tests__/authorize.test.js
```diff
@@ -0,0 +1,117 @@
...
+
+afterEach(() => {
+  jest.resetModules();
+});
+
+describe('adyen component authorize function', () => {
```
KadoBOT, line 16: Use just a name, to make it easier to identify tests
```suggestion
describe('Authorize', () => {
```

src/cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/middlewares/__tests__/authorize.test.js
```diff
@@ -0,0 +1,117 @@
...
+  it('return with appropriate mesages when create payment request fails', () => {
```
KadoBOT, line 17: for this assertion and the others, use the name convention:
```js
it('should return...')
```

src/cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/middlewares/__tests__/authorize.test.js
```diff
@@ -0,0 +1,117 @@
...
+    } = require('*/cartridge/scripts/adyenCheckout');
+    createPaymentRequest.mockImplementation(() => ({
+      error: {},
+    }));
+    const authorizeResult = authorize(
+      '15',
+      currentBasket.createPaymentInstrument(),
+      'mockedPaymentProcessor',
+    );
+    expect(authorizeResult).toMatchSnapshot();
```
KadoBOT, line 29: If you check the snapshot, the part with the mocks looks like this:
```
    "setCreditCardExpirationMonth": [MockFunction],
    "setCreditCardExpirationYear": [MockFunction],
    "setCreditCardNumber": [MockFunction],
    "setCreditCardToken": [MockFunction],
    "setCreditCardType": [MockFunction],
```
It doesnt matter if the mock was called or not, it won't change the snapshot. If you do care about the `authorizeResult`, you keep it. In addition you can have  something like this:
```js
expect(setCreditCardType.mock.calls).toMatchSnapshot()
```

src/cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/middlewares/handle.js
```diff
@@ -0,0 +1,51 @@
...
+    });
+    const paymentInstrument = currentBasket.createPaymentInstrument(
+      constants.METHOD_ADYEN_COMPONENT,
+      currentBasket.totalGrossPrice,
+    );
+    paymentInstrument.custom.adyenPaymentData = paymentInformation.stateData;
+    paymentInstrument.custom.adyenPaymentMethod =
+      paymentInformation.adyenPaymentMethod;
+
+    if (paymentInformation.isCreditCard) {
```
KadoBOT, line 22: this looks like a method. can you confirm? otherwise, this will always be truthy.

src/cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/middlewares/__tests__/authorize.test.js
zaiddreakh: I only added that part so the test will pass. Without them it complains.
Not sure I'm following what you mean with checking setCreditCardType

src/cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/middlewares/handle.js
zaiddreakh: I checked, it does return false when it's not a creditCard

src/cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/middlewares/__tests__/authorize.test.js
KadoBOT: What I mean is that this is not enough to know if the mock functions were called or not. They are always a `[MockFunction]`.
But if you test the snapshot of the call, you know if it's an empty array (not called) or an array with the arguments it was given when called.
i.e.:
```js
// snapshot (not called)
Array []
```
```js
// snapshot (called)
Array [
  Array [
    "/threeds2/adyen3ds2",
    Object {
      "environment": "test",
      "locale": "nl_NL",
      "originKey": "mocked_origin_key",
      "resultCode": "Authorised",
      "token3ds2": "mocked_token",
    },
  ],
]
```

PR #346, opened 2020-08-16
Unit tests for pos scripts

zaiddreakh:
## Tested scenarios

**Fixed issue**:


src/cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/middlewares/pos_authorize.js
```diff
@@ -0,0 +1,59 @@
+const server = require('server');
+const Resource = require('dw/web/Resource');
+const Transaction = require('dw/system/Transaction');
+const Logger = require('dw/system/Logger');
+
+/**
+ * Authorize
+ */
+function pos_authorize(orderNumber, paymentInstrument, paymentProcessor) {
```
KadoBOT, line 9: I have a question about this. Is this really a middleware? Because the pattern looks a bit different. But maybe it's just salesforce.

PR #357, opened 2020-09-03
Try..catch components, signature 3D, checkoutUrl

maassenbas:
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/controllers/Adyen.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/controllers/Adyen.js
@@ -1,5 +1,6 @@
 const server = require('server');
 const URLUtils = require('dw/web/URLUtils');
+const Logger = require('dw/system/Logger');
 const consentTracking = require('*/cartridge/scripts/middleware/consentTracking');
 const csrfProtection = require('*/cartridge/scripts/middleware/csrf');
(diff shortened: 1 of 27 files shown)


src/cartridges/int_adyen_SFRA/cartridge/controllers/Adyen.js
```diff
@@ -1,5 +1,6 @@
 const server = require('server');
 const URLUtils = require('dw/web/URLUtils');
+const Logger = require('dw/system/Logger');
```
KadoBOT, line 3: Is this being used? I see that you've added the Logger import, but I don't see another change using it.

src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/adyen3d.js
```diff
@@ -0,0 +1,30 @@
+const URLUtils = require('dw/web/URLUtils');
```
KadoBOT, line 1: can you write a unit test for this change?

KadoBOT (changes requested):
Can you also check the failing tests? You might need to update the snapshots.

KadoBOT (approved):

KadoBOT (approved):

zaiddreakh (approved):

PR #361, opened 2020-09-09
fix: wrap order cancellation in transaction

BilalDev:
## Summary
- On our store (lesgeorgettes.com) we have some issue while getting the event CAPTURE_FAILED. The order is not cancel.
- We need to wrap the order cancellation (undo cancellation too) to a transaction.

### Tested scenarios without this fix
- Create an order
- Simulate CAPTURE_FAILED event
- Check on BM, you should see the order is not cancel

### Tested scenarios **with** this fix
- Create an order
- Simulate CAPTURE_FAILED event
- Check on BM, you should see the order is cancel

### DIFF
--- a/src/cartridges/int_adyen_overlay/cartridge/scripts/handleCustomObject.js
+++ b/src/cartridges/int_adyen_overlay/cartridge/scripts/handleCustomObject.js
@@ -173,7 +173,9 @@ function handle(customObj) {
         if (customObj.custom.success === 'true') {
           order.setPaymentStatus(Order.PAYMENT_STATUS_NOTPAID);
           order.setExportStatus(Order.EXPORT_STATUS_NOTEXPORTED);
-          OrderMgr.cancelOrder(order);
+          Transaction.wrap(() => {
+            OrderMgr.cancelOrder(order);
+          });
         }
         Logger.getLogger('Adyen', 'adyen').info(
           'Capture Failed for order {0}',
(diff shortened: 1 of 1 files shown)


maassenbas:
Hi @BilalDev , 

The handleCustomObject should already be inside the Transaction.wrap in the notifications.js file, see here: 
https://github.com/Adyen/adyen-salesforce-commerce-cloud/blob/develop/src/cartridges/int_adyen_overlay/cartridge/scripts/job/notifications.js#L41-L43

This should make sure that the order gets cancelled correctly when handling the custom object.
Could you verify that on your end? 

Thanks,
Bas Maassen
Adyen

BilalDev:
Hi @maassenbas,

I think it does not work well. We do not have this issue anymore on our store since this fix.

Also, why are you wrapping the failOrder on [handleCustomObject.js](https://github.com/Adyen/adyen-salesforce-commerce-cloud/blob/63fd83f9b755eb8105451c420b934807f11ce167/src/cartridges/int_adyen_overlay/cartridge/scripts/handleCustomObject.js#L202) ?

Cheers,
Bilal

maassenbas:
Hi @BilalDev , 

I will test it on my sandbox environment and let you know my findings. 
I'll keep you updated. 

Cheers,
Bas

maassenbas:
Hi @BilalDev , 

I tested it on my environment and noticed that the Transaction.wrap() is not required to execute the cancelOrder() and undoCancelOrder(). 
This because these actions are already included in the Transaction.wrap() in the notifications.js file. 

Could you please verify on your end why this Transaction.wrap() is not working for you? 
Of course you can keep the newly added Transaction.wrap() on your end, but I will not merge it into the Adyen cartridge. 

Thanks,
Bas

BilalDev:
Hi,

Weird it doesn’t work for us. Why wrapping failOrder then ?

Cheers,
Bilal

maassenbas:
Hi @BilalDev , 

We can also remove the wrap around the failOrder indeed. 
Will update that in our new release.

Regards,
Bas

PR #366, opened 2020-09-17
refactor redirect

KadoBOT:


src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/redirect/signature.js
```diff
@@ -0,0 +1,42 @@
...
+  res.redirect(
+    URLUtils.url(
+      'Checkout-Begin',
+      'stage',
+      'payment',
+      'paymentError',
+      Resource.msg('error.payment.not.valid', 'checkout', null),
+    ),
+  );
+  return next();
```
maassenbas, line 23: Can we reuse this code in multiple controllers? I think this is used more often when a payment gets refused for example.

KadoBOT: we can and we should. let's plan that in a future stage.

maassenbas (approved):

PR #368, opened 2020-09-19
POS Terminal Empty Check

theShortJ:
Added empty JSON check for the POS terminal variable. It was causing the page to break.

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/getPaymentMethod/payment.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/getPaymentMethod/payment.js
@@ -28,7 +28,7 @@ function handlePaymentMethod({ req, res, next }) {
     AdyenPaymentMethods: response,
     ImagePath: adyenURL,
     AdyenDescriptions: paymentMethodDescriptions,
-    AdyenConnectedTerminals: JSON.stringify(connectedTerminals) == "{}" ? '' : JSON.parse(connectedTerminals)
+    AdyenConnectedTerminals: JSON.stringify(connectedTerminals) === "{}" ? '' : JSON.parse(connectedTerminals)
   };
 
   if (AdyenHelper.getCreditCardInstallments()) {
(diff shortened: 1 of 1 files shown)


maassenbas:
Hi @theShortJ , 

Thanks for you PR. 
We will verify it and add it to the code.

Kind regards,
Bas Maassen
Adyen

msilvagarcia (changes requested):
Hey there,

I believe there are better ways of checking for an empty object than using `JSON.stringify`.

Other than that, seems like a good idea.

Kind regards,
Marcos
Adyen

OleksiiKravchenkoOSF:
Hello, guys @maassenbas @msilvagarcia 

We also faced this issue and I think a more elegant solution will be if we just define the variable that going to be parsed as a string.
line 519 - `var connectedTerminals = "{}";` or `var connectedTerminals = JSON.stringify({})`

Thank you

maassenbas:
Hi @OleksiiKravchenkoOSF and @theShortJ , 

Basically, we already set the connectedTerminals to "{}" in the getConnectedTerminals() here: 
https://github.com/Adyen/adyen-salesforce-commerce-cloud/blob/develop/src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/getPaymentMethod/utils.js#L29 

I assume that will solve this issue.
Could you confirm that? 

Thanks,
Bas Maassen
Adyen

OleksiiKravchenkoOSF:
> Hi @OleksiiKravchenkoOSF and @theShortJ ,
> 
> Basically, we already set the connectedTerminals to "{}" in the getConnectedTerminals() here:
> https://github.com/Adyen/adyen-salesforce-commerce-cloud/blob/develop/src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/getPaymentMethod/utils.js#L29
> 
> I assume that will solve this issue.
> Could you confirm that?
> 
> Thanks,
> Bas Maassen
> Adyen

@maassenbas Yes, this issue won't happen with the latest version of Adyen. 
In my case, I have prev version like `Adyen Payment 20.1.0` where this issue occurs and probably we need to migrate to the latest one somewhen in the future. 

Thank you for your support!

maassenbas:
@OleksiiKravchenkoOSF Thanks for confirming! 
I will make sure this is covered in the upcoming release and close this PR then.

PR #370, opened 2020-09-22
Fix eslint warnings

KadoBOT:
Removed eslint rules that triggered warnings and fixed the subsequent errors


src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -11,7 +11,7 @@ const {
 } = require('./adyen_checkout/helpers');
 const { validateComponents } = require('./adyen_checkout/validateComponents');
 
-$('#dwfrm_billing').submit(function (e) {
+$('#dwfrm_billing').submit(function apiRequest(e) {
```
zaiddreakh, line 14: what's the purpose of this?

zaiddreakh (approved):

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
KadoBOT: It's a eslint rule. By having named functions, we facilitate our lives when debugging. Otherwise console will show "function anonymous bla bla bla"

PR #379, opened 2020-10-09
PW-2800 MB Way SFRA

zaiddreakh:
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
@@ -113,6 +113,35 @@ checkoutConfiguration.paymentMethodsConfiguration = {
       }
     },
   },
+  mbway: {
+    showPayButton: true,
+    onSubmit: (state, component) => {
+      $('#dwfrm_billing').trigger('submit');
+      assignPaymentMethodValue();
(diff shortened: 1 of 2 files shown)


src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -113,6 +113,30 @@ checkoutConfiguration.paymentMethodsConfiguration = {
...
+      $('#dwfrm_billing').trigger('submit');
+      assignPaymentMethodValue();
+      if (formErrorsExist) {
+        return false;
+      }
+      document.getElementById('component_mbway').querySelector('button').disabled = true;
+      paymentFromComponent(state.data, component);
+      document.querySelector('#adyenStateData').value = JSON.stringify(
+        state.data,
+      );
```
KadoBOT, line 128: Do you need to `return false`? Otherwise, this could be replaced with:
```suggestion
if (!formErrorsExist) {
      document.getElementById('component_mbway').querySelector('button').disabled = true;
      paymentFromComponent(state.data, component);
      document.querySelector('#adyenStateData').value = JSON.stringify(
        state.data,
      );
}
```

KadoBOT (approved):

maassenbas (approved):

PR #387, opened 2020-10-29
PW-3278 temporary fix for giropay

zaiddreakh:
## Summary

there's a bug in giropay from Checkout, it doesn't have the property "showPayButton", meaning it doesn't pick up the configuration "showPayButton: false" that we have for components.
For that reason, the pay button in giropay is always showing and is silently failing the form validation.
This is a temporary fix to hide the pay button and to pass the validation. Once Checkout fixes this bug in a new release, we will be getting rid of this temporary fix.


msilvagarcia (approved):

msilvagarcia:
Great explanation on why this is done. Good job! :)

maassenbas (approved):

Issue #343, opened 2020-08-13
Orders/payments to fail on open invoice payment options as Klarna and AfterpayTouch - error in the SFCC/Adyen cartridge


MaaGJ:
**Describe the bug**
Issue with open invoice data for orders with order-level discounts.

There is an issue with the line item price values sent to adyen, via the SFCC-adyen cartridge, for open invoices when an order has order-level discounts. The issue results in adyen failing the auth request and the customer is not able to use the open invoice payment method.
The problem is present in cartridge version 19.1 which is the version we are currently using. I have also looked at the code for the newest cartridge (19.2.2) and it looks like the problem is also present there, but I haven’t tested it.
The issue is in the following script, where line item data is collected and added to the adyen request object: 
•	/int_adyen_overlay/cartridge/scripts/adyenGetOpenInvoiceData.ds. 
The below is part of the JSON the script generates. To better illustrate what data is being sent to adyen I have inserted the commerce cloud functions that the script uses to fetch the required data:
![image](https://user-images.githubusercontent.com/69618455/90098003-902c8d00-dd37-11ea-8a52-26b23a726ac3.png)

Below are two examples of data being fetched from commerce cloud and sent to adyen via the cartridge. Scenario 1 works fine, but scenario 2 is where the issue occurs.
1.	Data for a line item without order-level discount:
![image](https://user-images.githubusercontent.com/69618455/90098035-a5a1b700-dd37-11ea-925b-a69077411d5d.png)
2.	Data for the same line item, but with a 10% order level discount:
![image](https://user-images.githubusercontent.com/69618455/90098066-b7835a00-dd37-11ea-8122-28efce0b690e.png)

As you can see in scenario 2, the amountExcludingTax (adjustedNetPrice) is higher than in scenario 1, which doesn't make sense since it includes 10% discount. One would assume that the adjustedNetPrice would be lower in case 2.
The reason for this is that commerce cloud includes the 10% discount in the adjustedTax (taxAmount), but the gross price does not include the discount. 

When commerce cloud calculates adjustedNetPrice, they take the adjustedGrossPrice 995 (not including 10% discount) and subtracts the adjustedTax 179,1 (including the 10% discount), resulting in an adjustedNetPrice of 815,9.
[...]

PR #388, opened 2020-11-03
PW-3313 tax calculation on gross site

zaiddreakh:
## Summary

taxing line items under order-level discounts in gross-based sites was causing problems for open invoice methods
https://github.com/Adyen/adyen-salesforce-commerce-cloud/issues/343

### DIFF
--- a/src/cartridges/int_adyen_overlay/cartridge/scripts/util/lineItemHelper.js
+++ b/src/cartridges/int_adyen_overlay/cartridge/scripts/util/lineItemHelper.js
@@ -57,7 +57,7 @@ const lineItemHelperObj = {
       || lineItem instanceof dw.order.ShippingLineItem
     ) {
       return AdyenHelper.getCurrencyValueForApi(lineItem.getAdjustedTax());
-    } if (lineItem instanceof dw.order.PriceAdjustment) {
+    } if (lineItem instanceof dw.order.PriceAdjustment && lineItem.getPromotion().getPromotionClass() !== 'ORDER') {
       return AdyenHelper.getCurrencyValueForApi(lineItem.tax);
     }
     return null;


msilvagarcia (approved):

maassenbas (approved):

MaaGJ:
I will review with my tech team and revert as soon as possible

MaaGJ:
Hi @zaiddreakh and @maassenbas, here's the feedback from my tech team

- Test that you can replicate the current issue. You replicate the issue by setting up an order level discount. Add a product to your basket triggering that discount and then try to pay with Klarna”
_Yes i can replicate the error with the current code version of Georg Jensen_

- “Implement the fix adyen has made in our cartridge setup”
_I have implemented the fix_

- “Try to replicate the issue again, and verify if their fix has solved the issue”
_I have tried to replicate the issue._
_The error is now an internal server error, from the Adyen side_

Response:
{{ {"status":500,"errorCode":"903","message":"Internal error: Result from Open Invoice service does not contain a redirect URL","errorType":"internal"}
}}


[merchantAccount.pdf](https://github.com/Adyen/adyen-salesforce-commerce-cloud/files/5517285/merchantAccount.pdf)

MaaGJ:
Hi @zaiddreakh  and @maassenbas 
Here's an update on your request to update the following line to “return 0” instead of “return null”? 
https://github.com/Adyen/adyen-salesforce-commerce-cloud/blob/8e43102683bbe547b973423a7c38c87638ffd39f/src/cartridges/int_adyen_overlay/cartridge/scripts/util/lineItemHelper.js#L63 


---
We tried adding “return 0” instead of “return null”, but now the function “getCurrencyValueForApi” in the adyenHelper.js script fails, because “0” does not have a currency. I tried fixing the new issue, and instead of returning 0, i returned “new Money(0, session.currency.currencyCode)”. This passes the getCurrencyValueForApi function, but the actual call to adyen still fails with internal status (error 500).

I have attached the generated json from the scenario with returning new Money(0, session.currency.currencyCode)

1 - Returning 0:
image

2 - Returning new Money(0, session.currency.currencyCode):
image

See attached
[adyenGetopenInvoiceData_failingRequet.json.docx](https://github.com/Adyen/adyen-salesforce-commerce-cloud/files/5848115/adyenGetopenInvoiceData_failingRequet.json.docx)

MaaGJ:
Hi @maassenbas 
perfect thanks, we have tested and verified that this now looks correct. 
We will implement the fix in the next sprint

PR #397, opened 2020-12-01
PW-3563 live endpoints in metadata

zaiddreakh:
## Summary

added a few more live endpoints that were missing in the services file in metadata
## Tested scenarios

**Fixed issue**:


maassenbas (changes requested):
Live endpoints do not seem to be correct. Could you verify?

maassenbas (approved):

KadoBOT (approved):

Issue #403, opened 2020-12-17
When PayPal dialog is closed before the content is loaded payment page is blocked - order is created but not failed


rumchoto:
**Describe the bug**
If you open the PayPal dialog from the button and then close it before the PayPal content is loaded you cannot proceed with checkout. Order is created but there is no event on close to fail it so you have to start over the whole checkout process.

**To Reproduce**
1. Add some product to cart
2. Go to checkout page
3. Fill needed data
4. Select PayPal payment method
5. Click PayPal button
6. Close the dialog before the PayPal login is loaded
7. Try to pay again

**Expected behavior**
On close of PayPal dialog is expected to be able to proceed the order either with the same payment method or different.
We expected the same behavior as on close of PayPal dialog when the PayPal content is loaded and onCancel function is called.

zaiddreakh:
Hey @rumchoto,

Thanks for pointing this out. We created an internal ticket and will update you once it's been fixed.

Regards,
Zaid
Adyen

rumchoto:
Hi @zaiddreakh,

Do you have any update on this issue?

Thanks,
Rumyana

zaiddreakh:
Hi @rumchoto,

This issue was fixed with this PR
https://github.com/Adyen/adyen-salesforce-commerce-cloud/pull/411

Regards,
Zaid
Adyen

rumchoto:
Hi @zaiddreakh,

Thank you for the update. 
We are looking forward to the next release.

Kind Regards,
Rumyana

PR #407, opened 2020-12-30
PW-3714 paypal cancel pop-up before it loads

zaiddreakh:
## Summary

When PayPal dialog is closed before the content is loaded payment page is blocked - order is created but not failed
https://github.com/Adyen/adyen-salesforce-commerce-cloud/issues/403

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
@@ -29,6 +29,7 @@ function getCardConfig() {
 function getPaypalConfig() {
   window.paypalTerminatedEarly = false;
   return {
+    showPayButton: true,
     environment: window.Configuration.environment,
     intent: 'capture',
     onSubmit: (state, component) => {
@@ -58,7 +59,7 @@ function getPaypalConfig() {
       document.querySelector('#showConfirmationForm').submit();
     },
     onClick: (data, actions) => {
-      if(window.paypalTerminatedEarly) {
+      if (window.paypalTerminatedEarly) {
         helpers.paymentFromComponent({ cancelTransaction: true });
         window.paypalTerminatedEarly = false;
         return actions.resolve();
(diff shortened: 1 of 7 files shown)


src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
```diff
@@ -39,21 +41,30 @@ function getPaypalConfig() {
       helpers.paymentFromComponent(state.data, component);
     },
     onCancel: (data, component) => {
+      window.paypalTerminatedEarly = false;
```
cyattilakiss, line 44: why is this line needed in all the callbacks?
I guess it would be enough only once the properties are defined like the `showPayButton: true,`.

src/cartridges/int_adyen_controllers/cartridge/js/pages/checkout/adyen-checkout.js
```diff
@@ -110,12 +122,15 @@ function initializeBillingEvents() {
           );
         },
         onCancel: (data, component) => {
+          paypalTerminatedEarly = false;
```
cyattilakiss, line 125: same here, maybe I just don't get the logic behind it :)

pavelkukov: Hi @cyattilakiss , a bit late to reply but...
Please, see the problem described in the original ticket.
https://github.com/Adyen/adyen-salesforce-commerce-cloud/issues/403

If the PayPal window is terminated by the user, there is no way to detect this in the checkout.

Best,
Pavel

zaiddreakh: Hi @pavelkukov,

this was fixed here https://github.com/Adyen/adyen-salesforce-commerce-cloud/pull/411

Regards,
Zaid
Adyen

pavelkukov: Thank you, @zaiddreakh !

pavelkukov: @zaiddreakh , I see the fix in develop branch only. When will be released?

PR #409, opened 2021-01-08
PW-3393 QR Codes

zaiddreakh:
## Summary

Support for QR code payment methods (swish, bcmc mobile, and weChatPayQR), and change of the cancellation flow for the function paymentFromComponent.
Once a QR code is rendered, if a different payment method was chosen then that QR code component is unmounted and a new component is created to replace it.

## Tested scenarios

- Happy flow
- shopper had QR code rendered then chose a different payment method
- QR code expired after countdown

**Fixed issue**:

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/qrCodeMethods.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/qrCodeMethods.js
@@ -1,0 +1,1 @@
+module.exports.qrCodeMethods = ['swish', 'wechatpayQR'];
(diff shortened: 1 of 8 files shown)


src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/qrCodeMethods.js
```diff
@@ -0,0 +1 @@
+module.exports.qrCodeMethods = ['swish', 'wechatpayQR'];
```
maassenbas, line 1: Could you also add/test bcmc_mobile here?

maassenbas (approved):

KadoBOT (approved):

PR #411, opened 2021-01-18
Paypal pop-up closed too early

zaiddreakh:
## Summary

When Paypal pop-up is closed before content loads it doesn't fire the event onCancel. Trying to checkout afterwards would result in an error
This PR handles this case where paypal is terminated early
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
@@ -27,6 +27,7 @@ function getCardConfig() {
 }
 
 function getPaypalConfig() {
+  window.paypalTerminatedEarly = false;
   return {
     showPayButton: true,
     environment: window.Configuration.environment,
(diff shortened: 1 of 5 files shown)


src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
```diff
@@ -49,18 +51,29 @@ function getPaypalConfig() {
       );
     },
     onError: (error, component) => {
+      window.paypalTerminatedEarly = false;
```
KadoBOT, line 54: Any specific reason why you need to set this value on the `window` object, instead of using the [store](./src/cartridges/int_adyen_SFRA/cartridge/store/index.js) ?

zaiddreakh: not really, was originally using it in the template files but eventually I didn't need them

maassenbas (dismissed):

KadoBOT (dismissed):
Can you modify the `store.js` to start the `store.paypalTerminatedEarly` with an initial value e.g `store.paypalTerminatedEarly = false;`? So you don't reference something `undefined`, besides that LGTM.

KadoBOT (approved):

maassenbas (approved):

PR #431, opened 2021-03-12
Checkout component to v3.18.2

zaiddreakh:
## Summary

Downgraded back to v3.18.2 since Swish doesn't work with version 20
## Tested scenarios

Ran e2e tests

**Fixed issue**:


cyattilakiss (approved):
Shouldn't the .map also be added/updated for this js?

src/cartridges/int_adyen_controllers/cartridge/js/pages/checkout/bundle.js

msilvagarcia (approved):

zaiddreakh:
the .map was never added to SG and it seems to work without it

PR #437, opened 2021-04-07
PW-4484: Possible values for paypal intent should only be capture and authorize

zaiddreakh:
## Summary

Possible values for paypal intent should be either capture or authorize
## Tested scenarios

**Fixed issue**:


maassenbas (approved):

mchetani:
Is there any timeline on merging this? or make it available for use?

PR #442, opened 2021-04-14
PW-4512: Fixes for issues found in test

zaiddreakh:
## Tested scenarios

**Fixed issue**:


src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/getPaymentMethod/payment.js
```diff
@@ -33,7 +33,7 @@ function handlePaymentMethod({ req, res, next }) {
     ImagePath: adyenURL,
     AdyenDescriptions: paymentMethodDescriptions,
     AdyenConnectedTerminals: JSON.parse(connectedTerminals),
-    amount: { value: paymentAmount, currency },
+    amount: { value: parseFloat(paymentAmount), currency },
```
cyattilakiss, line 36: Where is this amount used? In case it's is used for the frontend components as well as a config field you need to use minor units.

src/cartridges/int_adyen_controllers/cartridge/controllers/Adyen.js
```diff
@@ -408,7 +408,7 @@ function getPaymentMethods(cart, customer) {
     adyenConnectedTerminals: connectedTerminals,
     ImagePath: adyenURL,
     AdyenDescriptions: paymentMethodDescriptions,
-    amount: { value: paymentAmount, currency: currency },
+    amount: { value: parseFloat(paymentAmount), currency: currency },
```
cyattilakiss, line 411: Where is this amount used? In case it's is used for the frontend components as well as a config field you need to use minor units.

cyattilakiss (commented):
Which version of the checkout API is used?

zaiddreakh:
> Which version of the checkout API is used?

version 65

src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/getPaymentMethod/payment.js
zaiddreakh: @cyattilakiss yes it is used in the frontend. But I tried it in google pay and it works fine. Maybe minor units are required starting from a newer version?

cyattilakiss: it is required from the beginning. It won't always break the implementation but for example Apple pay would be immediately broken because of it and it can also cause problems in other components that we are not aware at the moment :(

cyattilakiss:
> > Which version of the checkout API is used?
> 
> version 65

@zaiddreakh I was asking it because I see you create the configuration objects on the frontend for example for google pay while it is returned in the checkout API but I think only starting with v66 so for now it's good like this :)

maassenbas (approved):

maassenbas (approved):

msilvagarcia (dismissed):

maassenbas (approved):

msilvagarcia (approved):

PR #468, opened 2021-06-02
Update Recurring settings

wboereboom:
## Summary
removed recurring enabled from cartridge 
& introduced proper values for recurring payments

### DIFF
--- a/src/cartridges/int_adyen_overlay/cartridge/scripts/util/adyenHelper.js
+++ b/src/cartridges/int_adyen_overlay/cartridge/scripts/util/adyenHelper.js
@@ -159,6 +159,10 @@ var adyenHelperObj = {
     };
   },
 
+  getAdyenRecurringPaymentsEnabled() {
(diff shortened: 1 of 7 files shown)


src/cartridges/int_adyen_overlay/cartridge/scripts/util/adyenHelper.js
```diff
@@ -171,11 +159,7 @@ var adyenHelperObj = {
     };
   },
 
-  getAdyenRecurringEnabled() {
-    return adyenHelperObj.getCustomPreference('AdyenRecurringEnabled');
-  },
-
-  getAdyenOneClickEnabled() {
```
maassenbas, line 178: Should we keep the "getAdyenOneClickEnabled" and delete the "getAdyenRecurringPaymentsEnabled" instead? This way it is more consistent with the configuration as the name in the Custom Preference is `AdyenOneClickEnabled`. What do you think?

maassenbas (approved):

zaiddreakh (approved):

PR #474, opened 2021-06-17
renamed adyenCheckout functions to match API endpoints

wboereboom:
## Summary
renamed adyenCheckout functions to match API endpoints for doPaymentsCall and doPaymentsDetailsCall


wboereboom:
I did not refactor `createPaymentRequest` in adyenCheckout.js on purpose as it does not directly relate to an API call. Interested to hear your thoughts!

zaiddreakh:
> I did not refactor `createPaymentRequest` in adyenCheckout.js on purpose as it does not directly relate to an API call. Interested to hear your thoughts!
you mean refactor to "createPaymentsRequest"? it's not a huge deal for me, but we do say "payments request" when we speak about it to each other, rather than "payment request", so it would make sense if you changed it.

zaiddreakh (approved):

maassenbas (approved):

PR #477, opened 2021-06-28
Pw 4996/update component sg

wboereboom:
## Summary
Update the component version to latest for SG and fix all breaking changes this causes. OriginKey is now deprecated in favor of clientKey


cyattilakiss (approved):

zaiddreakh (approved):

cyattilakiss:
@wboereboom are you going to remove the rest of the origin key related functionality in the next major release? (like: https://github.com/Adyen/adyen-salesforce-commerce-cloud/blob/20.1.3/src/cartridges/int_adyen_overlay/cartridge/scripts/adyenGetOriginKey.js ) Would it be a good idea to tag them as at deprecated in this PR?

wboereboom:
We already made this change for SFRA, and if you check develop you'll see that I already removed the adyenGetOriginKey.js class.

We'll have to at least update docs when we release this due to the necessity to include the clientKey in the cartridge configuration.

PR #499, opened 2021-07-06
Pw 5065/fix sfra checkout v67

wboereboom:
## Summary
After update to v67, a few breaking changes were introduced

No redirect object in the /payments response
No paymentData in the /payments response, which we used to use for hash
Test several payment flows and potentially add issues in the ticket (Redirect, 3DS1, 3DS2, Voucher, Pending)

## Tested scenarios
CreditCard 3DS1+2
iDeal
PayPal
Swish
Amazon Pay

### DIFF
--- a/src/cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.js
+++ b/src/cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.js
@@ -167,8 +167,8 @@ function doPaymentsCall(order, paymentInstrument, paymentRequest) {
     }
 
     paymentResponse.fullResponse = responseObject;
-    paymentResponse.redirectObject = responseObject.redirect
-      ? responseObject.redirect
+    paymentResponse.redirectObject = responseObject.action
+      ? responseObject.action
       : '';
     paymentResponse.resultCode = responseObject.resultCode;
     paymentResponse.pspReference = responseObject.pspReference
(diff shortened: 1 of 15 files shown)


src/cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.js
```diff
@@ -196,20 +196,14 @@ function doPaymentsCall(order, paymentInstrument, paymentRequest) {
...
       paymentResponse.threeDS2 = true;
-      let token3ds2;
-      if (responseObject.authentication['threeds2.fingerprintToken']) {
-        token3ds2 = responseObject.authentication['threeds2.fingerprintToken'];
-      } else if (responseObject.authentication['threeds2.challengeToken']) {
-        token3ds2 = responseObject.authentication['threeds2.challengeToken'];
-      }
-      paymentResponse.token3ds2 = token3ds2;
+
+      paymentResponse.token3ds2 = responseObject.action.token;
```
cyattilakiss, line 200: Is the `token3ds2` still used later? In case the action is already on the frontend I would suggest to let the component handle the `action` instead of "manually" handling the tokens with the threeDS2 components. 

I guess you have another ticket to actually do what I just suggested so if the case is that you want to tackle this in a separate PR pls ignore this comment :)

cyattilakiss (approved):

src/cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.js
wboereboom: This token is indeed used later  in the redirect, for which we still use a form post for now rather than the action. 
Indeed the story which is now on hold does away with this and will use the action rather than parse these kinds of fields manually (for both 3ds2 as well as 3ds1, redirect, etc.

zaiddreakh (approved):

src/cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.js
cyattilakiss: thanks for the elaborate answer! 👍

zaiddreakh (approved):

maassenbas (approved):

PR #504, opened 2021-07-12
Pw 4727/order confirm form

wboereboom:
## Summary
Order-confirm' route changed from a GET call to a POST call, throwing params don't match.


zaiddreakh (approved):
Looks good but to do we want to merge it already or wait until we figure out how to get the SFRA version?

wboereboom:
Good question... Let's see if we get a response from SF before too soon, otherwise I'd say let's merge it. especially as this PR fixes 3DS1.

maassenbas (approved):

PR #525, opened 2021-08-17
PW-5212 Amazon pay configuration update

zaiddreakh:
## Summary

Add Locale in Component config
Public Amazon key and Merchant Id can be retrieved from the /paymentMethods
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/renderGenericComponent.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/renderGenericComponent.js
@@ -108,7 +108,7 @@ module.exports.renderGenericComponent = async function renderGenericComponent() 
       data.AdyenPaymentMethods;
     setCheckoutConfiguration(data);
     const amazonpayConfig = data.AdyenPaymentMethods.paymentMethods.find(
-      (elm) => elm.type === 'amazonpay',
+      (paymentMethod) => paymentMethod.type === 'amazonpay',
     ).configuration;
     store.checkoutConfiguration.paymentMethodsConfiguration.amazonpay.configuration = amazonpayConfig; // eslint-disable-line max-len
     store.checkout = new AdyenCheckout(store.checkoutConfiguration);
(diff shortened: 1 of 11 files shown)


src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/renderGenericComponent.js
```diff
@@ -107,6 +107,10 @@ module.exports.renderGenericComponent = async function renderGenericComponent()
     store.checkoutConfiguration.paymentMethodsResponse =
       data.AdyenPaymentMethods;
     setCheckoutConfiguration(data);
+    const amazonpayConfig = data.AdyenPaymentMethods.paymentMethods.find(
```
wboereboom, line 110: If this configuration object in the `/paymentMethods` response applies to multiple payment methods: Can we make this generic so this will work with all payment methods that contain configuration data?

src/cartridges/int_adyen_controllers/cartridge/js/pages/checkout/adyen-checkout.js
```diff
@@ -670,11 +671,8 @@ function getAmazonpayConfig() {
       countryCode: paymentMethodsResponse.shippingAddress.country,
       phoneNumber: paymentMethodsResponse.shippingAddress.phone,
     },
-    configuration: {
-      merchantId: window.amazonMerchantID,
-      storeId: window.amazonStoreID,
-      publicKeyId: window.amazonPublicKeyID,
-    },
+    configuration: window.getPaymentMethodsResponse.adyenPaymentMethods.paymentMethods.find(
```
wboereboom, line 674: Same comment about generic solution applies here

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/renderGenericComponent.js
wboereboom: N.B: Maybe a name like `paymentMethod` would be more descriptive than `elm`.

wboereboom (changes requested):
Nitpicking again. Looks very clear and concise.

wboereboom (approved):
🌵

wboereboom (approved):
🌟

maassenbas (approved):

PR #531, opened 2021-08-24
Pw 5098/prefill address details

wboereboom:
## Summary
Prefill address details for all payment methods


maassenbas (dismissed):

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/renderPaymentMethod.js
```diff
@@ -10,13 +10,40 @@ function getFallback(paymentMethod) {
   return fallback[paymentMethod];
 }
 
+function getPersonalDetails() {
```
zaiddreakh, line 13: are there cases/PMs where we have to pass a state along with the address (U.S. states)?

zaiddreakh (approved):

PR #553, opened 2021-09-24
added Master Card

maassenbas:
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/src/cartridges/int_adyen_overlay/cartridge/scripts/util/adyenHelper.js
+++ b/src/cartridges/int_adyen_overlay/cartridge/scripts/util/adyenHelper.js
@@ -586,6 +586,8 @@ var adyenHelperObj = {
         case 'Master':
         case 'MasterCard':
         case 'Mastercard':
+        case 'Master Card':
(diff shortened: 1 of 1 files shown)


src/cartridges/int_adyen_overlay/cartridge/scripts/util/adyenHelper.js
```diff
@@ -586,6 +586,8 @@ var adyenHelperObj = {
         case 'Master':
         case 'MasterCard':
         case 'Mastercard':
+        case 'Master Card':
+        case 'Master card':
```
wboereboom, line 590: What if we clean `cardType` first by downcasing it and removing all spaces? That would leave us with a single case (i.e. `mastercard`). Of course the other cased would need adjusting as well.

maassenbas:
Closing this PR. 
The function `getAdyenCardType` is not used anymore since we are using the `state.data` as initial object.

vmklaassens:
@maassenbas can you update from which version onwards we use the ```state.data```?

PR #579, opened 2021-11-09
PW-5755 Dynamically create showConfirmationForm

zaiddreakh:
## Summary

showConfirmationForm is created in the template after the closing tag for dwfrm_billing

This can cause problems because nested forms are not allowed in HTML
and it doesn't work on testcafe

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/helpers.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/helpers.js
@@ -93,14 +93,12 @@ function showValidation() {
 }
 
 function createShowConfirmationForm() {
-  if(document.querySelector("#showConfirmationForm")) {
+  if (document.querySelector('#showConfirmationForm')) {
     return;
   }
   const template = document.createElement('template');
-  const form = `<form method="post"
-        id="showConfirmationForm"
-        name="showConfirmationForm"
-        action="${ShowConfirmationPaymentFromComponent}">
+  // eslint-disable-next-line no-undef
+  const form = `<form method="post" id="showConfirmationForm" name="showConfirmationForm" action="${ShowConfirmationPaymentFromComponent}">
     <input type="hidden" id="additionalDetailsHidden" name="additionalDetailsHidden" value="null"/>
     <input type="hidden" id="merchantReference" name="merchantReference"/>
     <input type="hidden" id="result" name="result" value="null"/>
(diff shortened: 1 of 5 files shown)


src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/helpers.js
```diff
@@ -92,10 +92,27 @@ function showValidation() {
     : displayValidationErrors();
 }
 
+function createShowConfirmationForm() {
+  if (document.querySelector('#showConfirmationForm')) {
+    return;
+  }
+  const template = document.createElement('template');
+  // eslint-disable-next-line no-undef
+  const form = `<form method="post" id="showConfirmationForm" name="showConfirmationForm" action="${ShowConfirmationPaymentFromComponent}">
```
wboereboom, line 101: How does this variable get filled? Can we pass it as a parameter instead of having it just pop up here?

zaiddreakh: you mean `${ShowConfirmationPaymentFromComponent}`?
I tried passing `window.ShowConfirmation` but it didn't work

src/cartridges/int_adyen_SFRA/cartridge/templates/default/checkout/billing/adyenComponentForm.isml
```diff
@@ -21,6 +21,7 @@
       window.paymentFromComponentURL = "${URLUtils.https('Adyen-PaymentFromComponent')}";
       window.resultCodeAuthorised = '${AdyenConstants.RESULTCODES.AUTHORISED}';
       window.paymentsDetailsURL = "${URLUtils.https('Adyen-PaymentsDetails')}";
+      window.ShowConfirmationPaymentFromComponent = "${ShowConfirmationPaymentFromComponent}";
```
maassenbas, line 24: Why do we need this as `window.ShowConfirmationPaymentFromComponent`? Is it used somewhere else?

zaiddreakh: now after the review changes yes :P
Before that I was calling `"${ShowConfirmationPaymentFromComponent}"` directly but it still had to be defined in the template, otherwise it's undefined

maassenbas (approved):

wboereboom (approved):

PR #585, opened 2021-11-18
Pw 5771/my account sg

wboereboom:
## Summary
Account tests SG


tests/e2e/pages/AccountPageSFRA.js
```diff
@@ -29,6 +29,7 @@ export default class AccountPage {
 
     addCard = async (cardData) => {
         await t
+            .expect(Selector('.card-error').visible).notOk()
```
zaiddreakh, line 32: I see the use case for this now but in the future we will probably forget why we have an expect statement in a helper function and it will look weird

zaiddreakh (approved):

maassenbas (approved):

PR #591, opened 2021-11-25
Added .value in handleCustomObject

maassenbas:
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/src/cartridges/int_adyen_overlay/cartridge/scripts/handleCustomObject.js
+++ b/src/cartridges/int_adyen_overlay/cartridge/scripts/handleCustomObject.js
@@ -127,7 +127,7 @@ function handle(customObj) {
         if (customObj.custom.success === 'true') {
           const amountPaid = parseFloat(order.custom.Adyen_value) + parseFloat(customObj.custom.value);
           const totalAmount = adyenHelper.getCurrencyValueForApi(adyenPaymentInstrument.getPaymentTransaction().getAmount()).value;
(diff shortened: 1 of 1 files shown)


wboereboom (approved):

src/cartridges/int_adyen_overlay/cartridge/scripts/handleCustomObject.js
```diff
@@ -165,7 +165,7 @@ function handle(customObj) {
             isAdyen
           ) {
             refusedHpp = true;
-          } else if (order.status === Order.ORDER_STATUS_FAILED) {
+          } else if (order.status.value === Order.ORDER_STATUS_FAILED) {
```
wboereboom, line 168: Should you do the same for line 251?

maassenbas: We could, but there it will already be converted to a String, so it should also work that way

wboereboom: Do we need to make it consistent over the entire class?

zaiddreakh (approved):

src/cartridges/int_adyen_overlay/cartridge/scripts/handleCustomObject.js
maassenbas: Let's do that during Notification refactoring

PR #608, opened 2021-12-27
Pw 6046 back button e2e

wboereboom:
## Summary
- E2E tests for back button during redirect 
- Bug fix in E2E back button checkout return stage. This will now always be 'shipping'


tests/e2e/fixtures/countriesEUR/NL.js
```diff
@@ -11,6 +12,17 @@ module.exports = (checkoutPage) => {
         await checkoutPage.expectSuccess();
     });
 
+    test.only('iDeal with restored cart success', async () => {
```
zaiddreakh, line 15: Looks great! Think you forgot to remove the `only` in this file

zaiddreakh (approved):

tests/e2e/pages/CheckoutPageSFRA.js
```diff
@@ -88,7 +88,7 @@ export default class CheckoutPageSFRA {
 
     setEmail = async () => {
         await t
-            .typeText(this.checkoutPageUserEmailInput, 'test@adyenTest.com');
+            .typeText(this.checkoutPageUserEmailInput, 'test@adyenTest.com', {replace: true});
```
maassenbas, line 91: Why is the `{replace: true}` needed?

maassenbas (dismissed):

tests/e2e/pages/CheckoutPageSFRA.js
wboereboom: When you return from the redirect this field might already be filled in (just like the rest of the form).
If I leave this out it will type the email in a field which already contains the email address

maassenbas (approved):

zaiddreakh (approved):

PR #622, opened 2022-01-31
PW-6143 Reviewing E2E tests

zaiddreakh:
## Summary

- Boleto is skipped because it doesn't seem to return an action anymore
- QRcode methods work but they look different between SG and SFRA
- some tests are still a bit flaky but now perform better with quarantine mode (e.g. Klarna, 3DS2 payment)
- restore cart still not supported in SG, and also flaky on SFRA (especially opening a new tab)
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "test": "jest ./src",
     "test:watch": "jest ./src --watch",
     "test:coverage": "jest ./src --coverage",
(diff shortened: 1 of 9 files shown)


wboereboom (approved):

maassenbas:
Regarding QR codes - instead of having different tests per environment, can we implement the QR feature in SG so it will be the same as SFRA?

maassenbas (approved):

PR #626, opened 2022-02-15
Pw 6132 sessions

wboereboom:
## Summary
sessions API can is used to replace the /paymentMethods call for us in SFCC.


src/cartridges/int_adyen_SFRA/cartridge/templates/default/checkout/billing/adyenGivingComponent.isml
```diff
@@ -1,10 +1,10 @@
 <isif condition="${pdict.adyen && pdict.adyen.adyenGivingAvailable}">
     <isset name="AdyenHelper" value="${require('*/cartridge/scripts/util/adyenHelper')}" scope="pdict"/>
     <link rel="stylesheet" type="text/css" href="${pdict.AdyenHelper.getCheckoutCSS()}"/>
+    <script src="${pdict.AdyenHelper.getCheckoutUrl()}"></script>
```
maassenbas, line 4: Should we use local script instead of external?

src/cartridges/int_adyen_overlay/cartridge/scripts/util/adyenHelper.js
```diff
@@ -442,10 +443,10 @@ var adyenHelperObj = {
     }
 
     const shopperDetails = {
-      firstName: args.order.getBillingAddress().firstName,
+      firstName: args.order.getDefaultShipment().getShippingAddress().firstName,
```
maassenbas, line 446: Why updating names to Shipping instead of Billing?

wboereboom: We need these details when creating a session object. At that stage we do not yet have any Billing information. That's why I changed it to shipping. 
Should we introduce a separate function for the session shopper details?
I guess it does make sense for this to be Billing.

src/cartridges/int_adyen_SFRA/cartridge/templates/default/checkout/billing/adyenGivingComponent.isml
wboereboom: Yes! I propose moving it to a separate PR so we can build on top of this one.

zaiddreakh (commented):
lint is unhappy with this line

zaiddreakh (changes requested):
lint is unhappy with this line

zaiddreakh (dismissed):

src/cartridges/int_adyen_overlay/cartridge/scripts/util/adyenHelper.js
maassenbas: I think it would make sense to have the Billing details as shopper details rather than Shipping.
Can we update the shopper details after the Session is created?

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyenCheckout.js
```diff
@@ -79,12 +80,6 @@ $('button[value="submit-payment"]').on('click', () => {
...
 
-const actionHandler = (action) => {
-  const checkout = new AdyenCheckout(store.checkoutConfiguration);
-  checkout.createFromAction(action).mount('#action-container');
-  $('#action-modal').modal({ backdrop: 'static', keyboard: false });
-};
-
 module.exports = {
   renderGenericComponent,
   actionHandler,
```
maassenbas, line 85: This `actionHandler` can be removed from the exports, right?

maassenbas (commented):
Had a few remarks, further looks good

src/cartridges/int_adyen_overlay/cartridge/scripts/util/adyenHelper.js
wboereboom: I'll change it so Shipping is used as a Fallback if Billing is not available. This way we use shipping to create the session, but we'll use billing for payments

zaiddreakh (approved):

maassenbas (approved):

PR #634, opened 2022-03-03
fix adyenGiving

wboereboom:
## Summary
update adyenGiving for SFRA so it is compatible with Checkout v5.x.x


maassenbas (approved):

maassenbas:
Do we also need to fix this for SG?

wboereboom:
Zaid has taken care of that in his PR!

zaiddreakh (approved):

PR #636, opened 2022-03-07
remove local bundle and css and replace with localized url based on c…

wboereboom:
## Summary
Removed local bundle file in favor of dynamic script tags.
Custom preference added to set shopper region so localized LIVE url can be used for increased performance.


metadata/site_import/meta/system-objecttype-extensions.xml
```diff
@@ -386,6 +386,24 @@
                 <externally-managed-flag>false</externally-managed-flag>
                 <min-length>0</min-length>
             </attribute-definition>
+            <attribute-definition attribute-id="Adyen_Frontend_Region">
+                <display-name xml:lang="x-default">Checkout frontend region</display-name>
+                <description xml:lang="x-default">The region to load your frontend resources (e.g. secured field iframes) from. Should be set to the region closest to where your shoppers are located.</description>
```
maassenbas, line 391: Should we refer here to the Web Components rather than Secured Fields?

metadata/site_import/meta/system-objecttype-extensions.xml
```diff
@@ -386,6 +386,24 @@
                 <externally-managed-flag>false</externally-managed-flag>
                 <min-length>0</min-length>
             </attribute-definition>
+            <attribute-definition attribute-id="Adyen_Frontend_Region">
+                <display-name xml:lang="x-default">Checkout frontend region</display-name>
```
maassenbas, line 390: Do we have to specify it is "frontend"? Or can we just call it "Checkout region" instead?

zaiddreakh (approved):

metadata/site_import/meta/system-objecttype-extensions.xml
wboereboom: I copied this from the way we do it at Magento. Happy to change the phrasing on this if we think we can make it more concise.

metadata/site_import/meta/system-objecttype-extensions.xml
wboereboom: I'm half assuming some merchants won't know our cartridge uses Web Components

maassenbas (approved):

astiskala:
I can't see in that where the `AdyenCheckout` environment is being changed to live-au or live-us (see [here](https://docs.adyen.com/online-payments/web-components#test-and-go-live) for context). Can I confirm that this is also part of this change?

wboereboom:
Hi @astiskala,
That's right. The URL's can be found [here](https://github.com/Adyen/adyen-salesforce-commerce-cloud/pull/636/files#diff-70ed78107472c508403579f2f88ed237827b6d3682f58932e569ca6f6d75392cR61)

Kind Regards,
Wouter
Adyen

astiskala:
@wboereboom Those look like the URLs for the resources (CSS + JS), not for the `environment` property (which is used to determine the endpoints for frontend API calls and SecuredFields).

wboereboom:
Hey @astiskala,

You're right. We were not aware of this setting. I'll create a ticket for it.
Thanks for bringing this under our attention.

PR #649, opened 2022-04-13
Add orderToken to order retrieval for SFRA

wboereboom:
## Summary
Update `OrderMgr.getOrder()` to always contain the orderToken when retrieving an order.

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/helpers.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/helpers.js
@@ -11,6 +11,15 @@ function assignPaymentMethodValue() {
   ).innerHTML;
 }
 
+function assignOrderFormValues(response) {
+  if (response.orderNo) {
+    document.querySelector('#merchantReference').value = response.orderNo;
+  }
+  if (response.orderToken) {
+    document.querySelector('#orderToken').value = response.orderToken;
+  }
+}
+
 /**
  * Makes an ajax call to the controller function PaymentFromComponent.
  * Used by certain payment methods like paypal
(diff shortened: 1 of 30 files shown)


src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/checkout/begin.js
```diff
@@ -9,7 +9,7 @@ const { updateSavedCards } = require('*/cartridge/scripts/updateSavedCards');
 
 function shouldRestoreBasket(cachedOrderNumber) {
   // restore cart if order number was cached
-  if (cachedOrderNumber !== undefined) {
+  if (cachedOrderNumber !== undefined && cachedOrderNumber !== null) {
```
zaiddreakh, line 12: can't we just check if`cachedOrderNumber` is truthy?

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/helpers.js
```diff
@@ -11,6 +11,15 @@ function assignPaymentMethodValue() {
   ).innerHTML;
 }
 
+function assignOrderFormValues(response) {
```
zaiddreakh, line 14: maybe we can find a better name for the function.. something like saveOrderIdentifiers?

src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/checkout/begin.js
wboereboom: Is order number always a String? If so then I'll change it to `if(cachedOrderNumber)`

maassenbas (dismissed):

maassenbas (dismissed):

zaiddreakh (dismissed):

zaiddreakh (approved):

maassenbas (approved):

PR #666, opened 2022-06-07
[PW-6717] Refactor e2e tests for Playwright

cenkiravul:
## Summary

**Fixed issue**:  PW-6717


maassenbas:
@cenkiravul , let's change this PR into `develop` instead of `main`

tests/playwright/sfcc.config.js
```diff
@@ -0,0 +1,128 @@
...
+    },
+
+    {
+      name: 'edge',
+      use: {
+        channel: 'msedge',
+      },
+    },
+
+    /* Test against mobile viewports. */
```
wboereboom, line 90: do we need to keep this commented code?

tests/playwright/package.json
```diff
@@ -0,0 +1,115 @@
+{
+  "devDependencies": {
+    "@playwright/test": "^1.21.0"
+  },
+  "name": "adyen_salesforce_e2e_tests",
+  "version": "1.0.0",
+  "description": "e2e tests for Adyen SalesForce Cartridges",
+  "main": "index.js",
+  "dependencies": {
```
wboereboom, line 9: should these be devDependencies?

wboereboom (changes requested):
Looks good. 2 small remarks :)

tests/playwright/sfcc.config.js
cenkiravul: It's a default config file, would be nice to keep them imho if we ever need to enable test cases for mobile viewports.

tests/playwright/package.json
cenkiravul: We only install playwright as dev dependency, the rest is added by playwright during installation. I don't have a strong opinion on this.

wboereboom (approved):

zaiddreakh (approved):

PR #670, opened 2022-07-11
Pw 6322/fix rounding errors

wboereboom:
## Summary
This fix introduces discount lines for discounted items in order to prevent half cents from being sent to checkout API lineitems.
It also rounds price and tax up/down to again prevent the half-cent issue.


zaiddreakh (approved):

shanikantsingh (commented):
@wboereboom there should also be some unit tests for these fixes/changes.

PR #681, opened 2022-08-11
remove stateData after succesful payment

wboereboom:


src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/showConfirmation.js
```diff
@@ -118,6 +118,7 @@ function showConfirmation(req, res, next) {
 
       // making sure Adyen_paymentMethod is populated before calling clearAdyenData()
```
zaiddreakh, line 119: there's a comment about clearing data but the command to do so wasn't there? that is strange

zaiddreakh (approved):

src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/showConfirmation.js
wboereboom: Yeah it's weird.. I'm going to delete this comment anyways. Looks like a dev thinking out loud rather than a helpful comment

maassenbas (approved):

PR #688, opened 2022-08-19
hide card brands inside card component

zaiddreakh:
Hide card brands inside card component following web components upgrade


wboereboom (approved):
Too bad they do not render properly. Good fix :)

maassenbas (approved):

PR #695, opened 2022-08-31
PW-7089: BCMC bug

zaiddreakh:
## Summary

It appears to be that following the web components upgrade, bcmc became independent from card component and needed its own component configurations.

**Fixed issue**:


wboereboom (approved):

wboereboom:
Should we add E2E for this new BCMC component?

maassenbas (approved):

wboereboom (approved):

PR #696, opened 2022-09-02
add klarna merchantdata to payment request

wboereboom:
## Summary
Set up additionalData field for Klarna click and collect feature

### DIFF
--- a/src/cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.js
+++ b/src/cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.js
@@ -27,6 +27,7 @@ const Logger = require('dw/system/Logger');
 const Resource = require('dw/web/Resource');
 const Order = require('dw/order/Order');
 const Transaction = require('dw/system/Transaction');
+const StringUtils = require('dw/util/StringUtils');
 
 /* Script Modules */
 const AdyenHelper = require('*/cartridge/scripts/util/adyenHelper');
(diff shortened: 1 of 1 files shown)


src/cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.js
```diff
@@ -112,6 +113,18 @@ function createPaymentRequest(args) {
...
+          shipping_method: order.getDefaultShipment().shippingMethod.displayName,
+          shipping_type: order.getDefaultShipment().shippingMethod.description,
+          first_name: address.firstName,
+          last_name: address.lastName,
+          street_address: `${address.address1} ${address.address2}`,
+          postal_code: address.postalCode,
+          city: address.city,
+          country: address.countryCode.value,
+        }
+        paymentRequest.additionalData['openinvoicedata.merchantData'] = StringUtils.encodeBase64(JSON.stringify(otherDeliveryAddress));
```
zaiddreakh, line 127: LGTM! maybe we can add a comment here to explain why the encoding is needed

zaiddreakh (approved):

src/cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.js
maassenbas: What if there is no shippingMethod, for example with pick up? It would throw an error here.

wboereboom: I have not been able to reproduce. The only store pickup shipping method I can create is one that costs 0. I think there should always be a shipping method.

zaiddreakh (dismissed):

src/cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.js
```diff
@@ -112,6 +113,19 @@ function createPaymentRequest(args) {
       args.addTaxPercentage = true;
       if(paymentRequest.paymentMethod.type.indexOf('klarna') > -1){
         args.addTaxPercentage = false;
+        const address = order.getBillingAddress();
+        const otherDeliveryAddress = {
+          shipping_method: order.getDefaultShipment().shippingMethod.displayName,
+          shipping_type: order.getDefaultShipment().shippingMethod.description,
```
amihajlovski, line 119: maybe we can define the `defaultShipment` variable same as address in lie 116 instead of invoking the service twice?

src/cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.js
```diff
@@ -112,6 +113,19 @@ function createPaymentRequest(args) {
       args.addTaxPercentage = true;
       if(paymentRequest.paymentMethod.type.indexOf('klarna') > -1){
```
amihajlovski, line 114: are we using formatting like prettier or something?

amihajlovski: also maybe in future this payment methods can be constants instead of using as strings, but not a big thing now.

maassenbas (approved):

zaiddreakh (approved):

PR #697, opened 2022-09-05
Pw 7122/bm config frontend logic

wboereboom:


zaiddreakh (approved):

src/cartridges/bm_adyen/cartridge/static/default/adyenSettings.js
```diff
@@ -1,41 +1,116 @@
...
+    let { value } = event.target;
+
+    // get checked boolean value for checkboxes
+    if (event.target.type === 'checkbox') {
+      value = event.target.checked;
+    }
+
+    // convert radio button strings to boolean if values are 'true' or 'false'
+    if (event.target.type === 'radio') {
+      if (event.target.value === 'true') {
```
maassenbas, line 51: We can also use `event.target.checked` instead to get a boolean value

maassenbas (approved):

PR #698, opened 2022-09-05
PW-7161 show discounted giftcard amount

zaiddreakh:
## Summary

refactoring of remaining amount container that shows up after applying a giftcard

**Fixed issue**:


src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/partialPayment.js
```diff
@@ -36,6 +36,12 @@ function makePartialPayment(req, res, next) {
       response.order.remainingAmount.currency,
     ).divide(100);
     response.remainingAmountFormatted = remainingAmount.toFormattedString();
+
+    const discountAmount = new Money(
+      response.amount.value,
+      response.amount.currency,
+    ).divide(100);
```
wboereboom, line 43: does this calculation check out for any currency? not just the ones using cents?

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
```diff
@@ -127,47 +128,50 @@ function removeGiftCard() {
...
-  const remainingAmountStart = document.createElement('div');
-  const remainingAmountEnd = document.createElement('div');
-  const cancelGiftCard = document.createElement('div');
-  const remainingAmountStartP = document.createElement('p');
-  const remainingAmountEndP = document.createElement('p');
-  const cancelGiftCardP = document.createElement('p');
-  const remainingAmountStartSpan = document.createElement('span');
-  const cancelGiftCardSpan = document.createElement('span');
-  const remainingAmountEndSpan = document.createElement('span');
+  Object.values(GIFTCARD_CONSTANTS).forEach((element) => {
```
maassenbas, line 133: I'm not sure if it is a good practice to add a loop for each constant. If we suddenly would add another constant that is related to giftcards, then another container will be created here. 
I would suggest to create a function instead that will create these containers and then add in the parameters which type of container to create.

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/constants.js
```diff
@@ -4,4 +4,9 @@ module.exports = {
   METHOD_ADYEN: 'Adyen',
   METHOD_ADYEN_POS: 'AdyenPOS',
   METHOD_ADYEN_COMPONENT: 'AdyenComponent',
+  GIFTCARD_CONSTANTS: {
+    GIFTCARDAMOUNT: 'Giftcard Amount',
+    REMAININGAMOUNT: 'Remaining Amount',
+    CANCELGIFTCARD: 'Cancel Giftcard?',
```
maassenbas, line 10: Should we use Resource files instead to also enable translations on the storefront?

zaiddreakh: good point

src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/partialPayment.js
zaiddreakh: do we already have a way to handle all currencies? or do we need to create something new
I tested with USD, Euro, and Indian rubees and they work fine
Japenese Yen however doesn't need this division

wboereboom: You could have a look at the `getCurrencyValueForApi` method in AdyenHelper

PR #706, opened 2022-10-04
Pw 7133/update component version to 5.26.0

wboereboom:
## Summary
Component version upgraded to 5.26.0


maassenbas:
Can we wait till version `5.27.0` and test that one accordingly?

PR #713, opened 2022-10-21
Bm config page

wboereboom:
## Summary
Custom Business manager configuration screen


zaiddreakh (approved):

src/cartridges/bm_adyen/cartridge/static/default/adyenSettings.js
```diff
@@ -0,0 +1,263 @@
...
+  const isValid = 'is-valid';
+  const isInvalid = 'is-invalid';
+  const adyenGivingBackground = document.querySelector(
+    '#fileDropBoxCharitybackground',
+  );
+  const adyenGivingLogo = document.querySelector('#fileDropBoxGivingLogo');
+  const params = 'resizable=yes,width=1000,height=500,left=100,top=100';
+
+  function settingChanged(key, value) {
+    const settingIndex = changedSettings.findIndex(
```
amihajlovski, line 24: instead of checking for index we can use `.some()` in future :)

src/cartridges/bm_adyen/cartridge/controllers/AdyenSettings.js
```diff
@@ -0,0 +1,83 @@
...
+        message,
+        error: true,
+      });
+      return next();
+    }
+
+    res.json({
+      success: true,
+      error: false,
+    });
```
amihajlovski, line 68: we could do something like this here...

`
const callResult = service.call(JSON.stringify({ merchantAccount }));
const isResultOk = callResult.isOk();
const responseData = {
    success: isResultOk,
    error: !isResultOk
};
const response = isResultOk ? responseData : { ...responseData, message: JSON.parse(callResult.getErrorMessage()) };
res.json(response);
return next();
`

src/cartridges/bm_adyen/cartridge/controllers/AdyenSettings.js
```diff
@@ -0,0 +1,83 @@
...
+  return next();
+});
+
+server.post('TestConnection', server.middleware.https, (req, res, next) => {
+  try {
+    const service = AdyenHelper.getService(
+      constants.SERVICE.CHECKOUTPAYMENTMETHODS,
+    );
+    if (!service) {
+      throw new Error('Could not do /paymentMethods call');
```
maassenbas, line 43: Perhaps indicate in the error log that this is the Test Connection button and not the /paymentMethods call in the checkout. For example: "Could not verify the configured merchant account and API key"

PR #729, opened 2022-11-10
PW-7446 cancel flow for applepay fix

zaiddreakh:
## Summary

- removed onError to solve cancel click on applepay
- refactored paymentFromComponent to solve eslint complexity warning
- removed applepay from blocked payment methods list

**Fixed issue**:

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
@@ -26,12 +26,13 @@ function getPaypalConfig() {
     showPayButton: true,
     environment: window.Configuration.environment,
     onSubmit: (state, component) => {
-      helpers.assignPaymentMethodValue();
-      document.querySelector('#adyenStateData').value = JSON.stringify(
-        store.selectedPayment.stateData,
-      );
-
-      helpers.paymentFromComponent(state.data, component);
+      setTimeout(() => {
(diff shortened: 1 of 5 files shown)


src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
```diff
@@ -26,12 +26,13 @@ function getPaypalConfig() {
...
-      );
-
-      helpers.paymentFromComponent(state.data, component);
+      setTimeout(() => {
+        helpers.assignPaymentMethodValue();
+        document.querySelector('#adyenStateData').value = JSON.stringify(
+          store.selectedPayment.stateData,
+        );
+        helpers.paymentFromComponent(state.data, component);
+      }, 100);
```
wboereboom, line 35: why 100? do we want to make this configurable?

wboereboom (commented):
1 question. Looks good otherwise :)

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
zaiddreakh: This seems to be a bug that I linked in the summary. Ideally we shouldn't have a setTimeout in the first place. I think the problem is the component's status is sometimes not set to `Ready` or takes a long time to do so.  Let me look further into it, I will try if we can change the component status manually

wboereboom (approved):

PR #748, opened 2022-11-21
Boolean type Custom Preferences fix

MilanMasta:
## Summary
When set to false `Boolean` type `Custom Preferences` returning `null` instead of `false`

**Fixed issue**:


cartridges/int_adyen_overlay/cartridge/scripts/util/adyenConfigs.js
```diff
@@ -4,7 +4,7 @@ var dwsystem = require('dw/system');
 var adyenCurrentSite = dwsystem.Site.getCurrent();
 function getCustomPreference(field) {
   var customPreference = null;
-  if (adyenCurrentSite && adyenCurrentSite.getCustomPreferenceValue(field)) {
+  if (adyenCurrentSite) {
```
MilanMasta, line 7: So the problem was that if we had a `field` ( `customPref `) that is `bool` type (for example `AdyenOneClickEnabled`) and it is set to `false`, it will skip this if statement because the second condition is not fulfilled and `customPreference` will be left on `null` and not `false` as it should.
Also it doesn't make to much sense to have the second condition because if we the `filed` name (`customPref`) doesn't exist it will ret null and it will set the `customPreference` to `null` anyways

zaiddreakh (approved):
Thanks for opening this PR and for the explanation! @MilanMasta

MilanMasta:
> Thanks for opening this PR and for the explanation! @MilanMasta

No problem :) @zaiddreakh

amihajlovski (approved):

PR #749, opened 2022-11-22
Pw 7284/multiple giftcard brands

amihajlovski:
## Summary

Allow shoppers to use different giftcard brands instead of only one.
New giftcard UI
Modified payment methods UI
Removed modal

## Test scenarios
1. Adding different giftcard brands
2. Removing giftcard
3. Persist giftcard data on refresh and canceled payment after adding
4. Add giftcard -> cancel -> add another giftcard brand
5. Add giftcard -> remove -> add another giftcard brand
6. Add giftcard -> go to home -> add another item to cart -> giftcard should be removed and message should be shown
7. Do payment and try adding giftcard to another payment


src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
```diff
@@ -133,6 +146,166 @@ function showGiftCardWarningMessage() {
   orderTotalSummaryEl.appendChild(alertContainer);
 }
 
+function attachGiftCardFormListeners() {
+  const giftCardSelect = document.querySelector('#giftCardSelect');
```
wboereboom, line 150: Can we move these selectors to their own method? We repeat the code to retrieve them (including string ID as argument) throughout this class. Might be more scalable if we have that in a single function that returns the selected UI elements as an object.

wboereboom (changes requested):

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
amihajlovski: Great tip! I will do that

zaiddreakh:
looks good @amihajlovski, can we move some of the giftcard functions to a separate file out of the checkoutConfiguration file? because they are too many and the file is quite big now

wboereboom (dismissed):

zaiddreakh (dismissed):

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/renderGiftcardComponent.js
```diff
@@ -0,0 +1,352 @@
...
+  const giftCardDiv = document.createElement('div');
+  giftCardDiv.classList.add('gift-card');
+
+  const brandContainer = document.createElement('div');
+  brandContainer.classList.add('brand-container');
+
+  const giftCardImg = document.createElement('img');
+  const giftCardImgSrc = `${imagePath}${giftCardData.brand}.png`;
+  giftCardImg.setAttribute('src', giftCardImgSrc);
+  giftCardImg.width = 40;
```
maassenbas, line 220: Can't we use the css class for this instead of hard coding width/height values?

amihajlovski: This is now resolved :)

zaiddreakh (dismissed):

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/renderGiftcardComponent.js
```diff
@@ -0,0 +1,351 @@
+const store = require('../../../../store');
+const constants = require('../constants');
+
+function showGiftCardWarningMessage() {
+  const alertContainer = document.createElement('div');
+  alertContainer.setAttribute('id', 'giftCardWarningMessage');
+  alertContainer.classList.add('alert', 'alert-warning', 'error-message');
+  alertContainer.style.display = 'block';
+  alertContainer.style.margin = '20px 0';
```
maassenbas, line 9: Can we move this to css as well?

maassenbas (approved):

zaiddreakh (approved):

PR #758, opened 2022-12-06
Pw 7462/confirmation email Site Genesis

zenit2001:
This PR resolves the issue of sending multiple emails when confirmation page is refreshed.
For payments completed directly from components (e.g. PayPal), it does the payment call when order is created, instead of every time the function is triggered.

### DIFF
--- a/src/cartridges/int_adyen_controllers/cartridge/controllers/Adyen.js
+++ b/src/cartridges/int_adyen_controllers/cartridge/controllers/Adyen.js
@@ -386,13 +386,13 @@ function showConfirmationPaymentFromComponent() {
     paymentInformation.get('paymentFromComponentStateData').value(),
   );
 
-  let finalResult;
+  let amazonPayResult;
 
   const hasStateData = stateData && stateData.details && stateData.paymentData;
 
   if (!hasStateData) {
     if (result && JSON.stringify(result).indexOf('amazonpay') > -1) {
-      finalResult = JSON.parse(result);
+      amazonPayResult = JSON.parse(result);
     } else {
       // The billing step is fulfilled, but order will be failed
       app.getForm('billing').object.fulfilled.value = true;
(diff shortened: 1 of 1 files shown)


amihajlovski (approved):

src/cartridges/int_adyen_controllers/cartridge/controllers/Adyen.js
```diff
@@ -426,17 +424,28 @@ function showConfirmationPaymentFromComponent() {
     adyenPaymentInstrument.custom.adyenPaymentData = null;
   });
 
-  finalResult = finalResult || adyenCheckout.doPaymentsDetailsCall(requestObject);
-
-  if (['Authorised', 'Pending', 'Received'].indexOf(finalResult.resultCode) > -1) {
+  if (order.status.value === Order.ORDER_STATUS_CREATED) {
+    finalResult = finalResult || adyenCheckout.doPaymentsDetailsCall(requestObject);
```
wboereboom, line 428: Can we introduce an additional value to get rid of `finalResult = finalResult`? I would like to suggest introducing `amazonPayResult` at line 389 and to introduce finalResult at line 426.
I think that would make this class a little less confusing

src/cartridges/int_adyen_controllers/cartridge/controllers/Adyen.js
```diff
@@ -426,17 +424,28 @@ function showConfirmationPaymentFromComponent() {
...
     Transaction.wrap(() => {
       AdyenHelper.savePaymentDetails(adyenPaymentInstrument, order, finalResult);
     });
-    OrderModel.submit(order);
     clearForms();
     app.getController('COSummary').ShowConfirmation(order);
     return {};
   }
+  // handles the refresh
+  else if (order.status.value === Order.ORDER_STATUS_CREATED || order.status.value === Order.ORDER_STATUS_NEW || order.status.value === Order.ORDER_STATUS_OPEN){
```
wboereboom, line 445:
```suggestion
  else if (
    [
      Order.ORDER_STATUS_CREATED, 
      Order.ORDER_STATUS_NEW,
      Order.ORDER_STATUS_OPEN,
      ].indexOf(order.status.value) > -1
      ) {
```

src/cartridges/int_adyen_controllers/cartridge/controllers/Adyen.js
zenit2001: Will do so!

src/cartridges/int_adyen_controllers/cartridge/controllers/Adyen.js
zenit2001: Suggestion added to last commit!

wboereboom (approved):

PR #759, opened 2022-12-07
[PW-7555] Maintain broken E2E tests

cenkiravul:
## Summary

- All tests that could be unskipped are unskipped and fixed
- Each product (SG, SFRA 5 and SFRA 6) has a seperate pipeline now, they are executed in parallel and each test run is expected to finish ~5 minutes
- The pipeline of each product category should be triggered when only a change is done to a certain product. E.g. A change to SG only should not trigger SFRA tests
- Introducing "Quick Coverage": If the PR is against develop or main, all tests will run, if not, only 10 tests which have `@quick` tag will be executed to save time.

**Fixed issue**:

### DIFF
--- a/tests/playwright/fixtures/BRL.spec.mjs
+++ b/tests/playwright/fixtures/BRL.spec.mjs
@@ -43,6 +43,7 @@ for (const environment of environments) {
       await checkoutPage.expectRefusal();
     });
 
+    // LPM needs to fix Boleto
     test.fixme('Boleto Success', async ({ page }) => {
       await new PresentToShoppers(page).doBoletoPayment();
       await checkoutPage.completeCheckout();
--- a/tests/playwright/fixtures/BRL.spec.mjs
+++ b/tests/playwright/fixtures/BRL.spec.mjs
@@ -15,7 +15,6 @@ let cards;
 
 for (const environment of environments) {
   test.describe.parallel(`${environment.name} BRL`, () => {
-    if (environment.name === 'SG') test.fixme();
 
     test.beforeEach(async ({ page }) => {
       await page.goto(`${environment.urlExtension}`);
--- a/tests/playwright/fixtures/BRL.spec.mjs
+++ b/tests/playwright/fixtures/BRL.spec.mjs
@@ -43,7 +43,7 @@ for (const environment of environments) {
       await checkoutPage.expectRefusal();
     });
 
-    // LPM needs to fix Boleto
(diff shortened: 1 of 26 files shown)


tests/playwright/pages/CheckoutPageSFRA5.mjs
```diff
@@ -85,16 +86,16 @@ export default class CheckoutPageSFRA5 {
...
   navigateToCheckout = async (locale) => {
     await this.page.goto(this.getCheckoutUrl(locale));
   };
 
-  goToCheckoutPageWithFullCart = async (locale) => {
-    await this.addProductToCart(locale);
-    await this.successMessage.waitFor({ visible: true, timeout: 10000 });
+  goToCheckoutPageWithFullCart = async (locale, itemCount = 1) => {
+    await this.addProductToCart(locale, itemCount);
+    await this.successMessage.waitFor({ visible: true, timeout: 15000 });
```
wboereboom, line 98: Can we move timeout to a variable for all CheckoutPages?

wboereboom (changes requested):

tests/playwright/fixtures/BRL.spec.mjs
```diff
@@ -43,6 +43,7 @@ for (const environment of environments) {
       await checkoutPage.expectRefusal();
     });
 
+    // LPM needs to fix Boleto
```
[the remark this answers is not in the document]
cenkiravul: This is removed, good call

tests/playwright/pages/CheckoutPageSFRA5.mjs
cenkiravul: Potentially we can, and I thought about it. However certain actions may need different timeouts, hence static usage. In test config file we have different timeouts set for actions, assertions and the total test run. The places we have overriden this are preventing early failures.

I agree that we can improve the whole approach to timeouts, especially after working with resourceful runners, but I'd rather handle it with care, after having run the tests with the current setting

wboereboom (approved):

tests/playwright/pages/CheckoutPageSFRA5.mjs
zaiddreakh: we can possibly set a variable for default timeout time, and unless passed a different value we use the default

zaiddreakh (approved):

amihajlovski (approved):

PR #801, opened 2023-01-06
[PW-7666] Sync payment instrument between OMS and SFCC

amihajlovski:
## Summary
This feature allows the OMS to have a clear separation between the payment method used as part of the order. 
If the shopper paid with a credit card then `CREDIT_CARD` type will be used as payment instrument, otherwise `AdyenComponent` will be set. 
For both cases same payment processor will be used `Adyen_Component`.


metadata/site_import/meta/system-objecttype-extensions.xml
```diff
@@ -120,6 +120,14 @@
                 <min-length>0</min-length>
                 <field-length>0</field-length>
             </attribute-definition>
+            <attribute-definition attribute-id="adyenMainPaymentInstrument">
+                <display-name xml:lang="x-default">adyenMainPaymentInstrument</display-name>
+                <description xml:lang="x-default">This setting is used to determine which payment method is used during checkout</description>
+                <type>text</type>
```
zaiddreakh, line 126: Since we're only accepting 2 values, maybe we can set the type to enum of strings and define those 2 values?

metadata/site_import/sites/RefArch/payment-methods.xml
```diff
@@ -12,4 +12,10 @@
         <enabled-flag>true</enabled-flag>
         <processor-id>Adyen_Component</processor-id>
```
zaiddreakh, line 13: in the description of adyen_component let's add a more clear description to differentiate it from credit cards. the description should say that this one is used for all alternative payment methods.

src/cartridges/int_adyen_overlay/cartridge/scripts/util/adyenHelper.js
```diff
@@ -520,6 +521,16 @@ var adyenHelperObj = {
...
   },
 
+  getOrderMainPaymentInstrumentType(order) {
+    let type = null;
+    collections.forEach(order.getPaymentInstruments(), (item) => {
+      if (item.custom.adyenMainPaymentInstrument) {
+        type = item.custom.adyenMainPaymentInstrument;
+      }
+    });
+    return type;
```
zaiddreakh, line 531: if type was null, would `order.getPaymentInstruments(adyenHelperObj.getOrderMainPaymentInstrumentType(order))[0];` throw any errors?

src/cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/middlewares/handle.js
```diff
@@ -1,52 +1,64 @@
...
   Transaction.wrap(() => {
-    collections.forEach(currentBasket.getPaymentInstruments(), (item) => {
-      currentBasket.removePaymentInstrument(item);
-    });
+    removeAllPaymentInstruments(currentBasket);
+
+    const paymentInstrumentType = paymentInformation.isCreditCard
+      ? PaymentInstrument.METHOD_CREDIT_CARD
+      : constants.METHOD_ADYEN_COMPONENT;
     const paymentInstrument = currentBasket.createPaymentInstrument(
```
zaiddreakh, line 45: let's add a unit test for this to make sure it chooses the right one when creating a payment instrument

zaiddreakh (changes requested):
Nicely done! left some small notes

metadata/site_import/sites/RefArch/payment-methods.xml
amihajlovski: Added descriptions.

metadata/site_import/meta/system-objecttype-extensions.xml
amihajlovski: Added this as an enum. Great suggestion 👍

src/cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/middlewares/handle.js
amihajlovski: Test added for checking payment type.

src/cartridges/int_adyen_SFRA/cartridge/scripts/hooks/payment/processor/middlewares/handle.js
```diff
@@ -1,52 +1,62 @@
...
+
+function removeAllPaymentInstruments(currentBasket) {
+  collections.forEach(currentBasket.getPaymentInstruments(), (item) => {
+    currentBasket.removePaymentInstrument(item);
+  });
+}
+
+function convertToSfccCardType(paymentInformation, paymentInstrument) {
+  const sfccCardType = !paymentInformation.creditCardToken
+    ? AdyenHelper.getSFCCCardType(paymentInformation.cardType)
```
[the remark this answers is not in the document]
amihajlovski: good point 👍 
resolved.

shanikantsingh (approved):

reejojoseph:
Hello @amihajlovski 

I think I found one issue with this change. I am not completely sure. But just wanted to check. As part of this change the payment method was changed to CREDIT_CARD from ADYEN_COMPONENT for credit card payments. The payment processor remains the same: **Adyen_Component**. So, when the Webhook notification hits SFCC from Adyen, there is condition specified on 

**handleCustomObject.js**:

`if ([constants.METHOD_ADYEN, constants.METHOD_ADYEN_POS, constants.METHOD_ADYEN_COMPONENT].indexOf(paymentInstruments[pi].paymentMethod) !== -1 || PaymentMgr.getPaymentMethod(paymentInstruments[pi].getPaymentMethod()).getPaymentProcessor().ID === 'ADYEN_CREDIT') {
            isAdyen = true;
            paymentInstruments[pi].paymentTransaction.custom.Adyen_log = customObj.custom.Adyen_log;
            adyenPaymentInstrument = paymentInstruments[pi];
          }`


In the IF condition, only the constants **constants.METHOD_ADYEN, constants.METHOD_ADYEN_POS, constants.METHOD_ADYEN_COMPONENT** (+ payment processor check is **ADYEN_CREDIT**). 

So, the question is, for credit card transactions, wont this IF clause always fail after this change?

In my case, the code is breaking at line: 106 where adyenPaymentInstrument is referenced as it doesn't have a null check.

Was there a work around on your side or its a miss?

Let me know if you need more details.

Many Thanks,
Reejo

amihajlovski:
Hello @reejojoseph 

Thanks for pointing this out. I took a look and we will add a fix in the next patch release that we plan soon.

Kind regards,
Aleksandar
Adyen

PR #831, opened 2023-02-16
PW-7784: rerender express component

shanikantsingh:
## Summary

**Fixed issue**:


zaiddreakh (commented):
about the Lodash dependency, is it worth having a new dependency just to use debounce? Maybe it's better we create our own debounce function without libraries?
Besides that it looks good to me!

zaiddreakh (approved):

zenit2001 (approved):

PR #835, opened 2023-02-23
PW-8078 Update API name for the field adyenPaymentMethod in SFCC

zaiddreakh:
Updating the adyenPaymentMethod property name for payment instrument so that it syncs with OMS

### DIFF
--- a/metadata/site_import/meta/system-objecttype-extensions.xml
+++ b/metadata/site_import/meta/system-objecttype-extensions.xml
@@ -92,6 +92,15 @@
                 <min-length>0</min-length>
                 <field-length>0</field-length>
             </attribute-definition>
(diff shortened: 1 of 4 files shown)


metadata/site_import/meta/system-objecttype-extensions.xml
```diff
@@ -92,6 +92,15 @@
                 <min-length>0</min-length>
                 <field-length>0</field-length>
             </attribute-definition>
+            <attribute-definition attribute-id="adyen_payment_Adyen_Payment_Method">
```
maassenbas, line 95: The attribute `adyen_payment_adyenPaymentMethod` can be removed then, right?

maassenbas (approved):

shanikantsingh (approved):

PR #902, opened 2023-04-13
Change value assigned to adyenPaymentMethod from innerHTML to textContent

kalinstanchev:
since innerHTML also takes the HTML and saves it in the order information. Issue found if you translate your frontend page with the browser in use which adds extra HTML tags like font and spans.
Changing the code form innerHTML to textContent it takes only the text inside the provided id tag in the JS. If you keep innerHTML some customers translate their front which adds extra HTML tags in, for example, Chrome adds 2 font HTML tags, even for text which can't be translated example Paypal.

## Summary

**Fixed issue**:


amihajlovski:
Hi @kalinstanchev 

Thanks for raising this. I will have to close this PR because its done in the build directory, but will create an issue and we will pick it up from there.

Kind Regards,
Aleksandar
Adyen

PR #914, opened 2023-05-22
HMAC Validation

zenit2001:
This PR introduces the option of using HMAC validation to validate notifications coming from Adyen in SFRA architecture.


src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/notify.js
```diff
@@ -1,20 +1,35 @@
...
   const notificationResult = handleNotify.notify(req.form);
-
-  if (notificationResult.success) {
+  if (notificationResult.success && hmacVerification){
+    Transaction.commit();
+    res.render('/notify');
+  }
+  else if (notificationResult.success) {
     Transaction.commit();
     res.render('/notify');
```
amihajlovski, line 34: The if and else/if statement do the same..
What is the purpose of this check?

zenit2001: The first if statement is executed if there is an hmacVerification needed meaning that hmac validation is enabled, while the else if block will be executed when hmac validation is disabled

src/cartridges/int_adyen_overlay/cartridge/scripts/checkNotificationAuth.js
```diff
@@ -40,6 +41,30 @@ function check(request) {
   );
 }
 
+function compareHmac(hmacSignature, merchantSignature){
+  let bitwiseComparison;
+  if (hmacSignature.length !== merchantSignature.length){
+    return false;
+  }
+  for (let i = 0; i <= (hmacSignature.length - 1); i++) {
```
amihajlovski, line 49: you can use `i < hmacSignature.length`

src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/notify.js
```diff
@@ -1,20 +1,35 @@
 const Transaction = require('dw/system/Transaction');
 const checkAuth = require('*/cartridge/scripts/checkNotificationAuth');
 const handleNotify = require('*/cartridge/scripts/handleNotify');
+const AdyenConfigs = require('*/cartridge/scripts/util/adyenConfigs');
 
 /**
  * Called by Adyen to update status of payments. It should always display [accepted] when finished.
  */
+
+function handleHmacVerification(req){
```
amihajlovski, line 10: This function might return undefined instead of boolean. 
Instead maybe you can use
```
if (!hmacKey) {
return false;
}
return checkAuth.validateHmacSignature(req);
```

src/cartridges/int_adyen_overlay/cartridge/scripts/checkNotificationAuth.js
```diff
@@ -40,6 +41,30 @@ function check(request) {
...
+  return bitwiseComparison === 0;
+};
+
+function validateHmacSignature(request){
+  const notificationData = request.form;
+  const hmacSignature = notificationData['additionalData.hmacSignature'];
+  const merchantSignature = AuthenticationUtils.calculateHmacSignature(request);
+  // Checking for timing attacks
+  if (compareHmac(hmacSignature, merchantSignature)){
+    AdyenLogs.info_log(`Successfully compared the HMAC Signatures, notification is coming from Adyen`);
```
zaiddreakh, line 61: I don't think there is a need to log anything when it's successful. Instead we can add a log when it errors to say that there is a mismatch

zaiddreakh (commented):
Looks good to me, just a small comment

src/cartridges/int_adyen_overlay/cartridge/scripts/checkNotificationAuth.js
zenit2001: Updated it to the error scenario

zaiddreakh (approved):

amihajlovski (approved):

zaiddreakh (approved):

PR #915, opened 2023-05-24
Enable klarna widget

shanikantsingh:
Use Klarna inline Widget in place of redirect.

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
@@ -311,21 +311,25 @@ function getApplePayConfig() {
 }
 
 function getKlarnaConfig() {
-  return {
-    showPayButton: true,
-    useKlarnaWidget: true,
-    onSubmit: (state, component) => {
-      helpers.assignPaymentMethodValue();
-      helpers.paymentFromComponent(state.data, component);
-    },
-    onAdditionalDetails: (state) => {
-      document.querySelector('#additionalDetailsHidden').value = JSON.stringify(
-        state.data,
-      );
-      document.querySelector('#showConfirmationForm').submit();
-    },
+  const {klarnaWidgetEnabled} = window;
+  if (klarnaWidgetEnabled){
+    return {
+      showPayButton: true,
+      useKlarnaWidget: true,
+      onSubmit: (state, component) => {
+        helpers.assignPaymentMethodValue();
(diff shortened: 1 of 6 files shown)


amihajlovski:
Please update the PR title so we don't need to do changes in the release notes. 
And also update the labels.

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/helpers.js
```diff
@@ -73,7 +73,7 @@ function displaySelectedMethod(type) {
   resetPaymentMethod();
 
   document.querySelector('button[value="submit-payment"]').disabled =
-    ['paypal', 'paywithgoogle', 'googlepay', 'amazonpay', 'applepay'].indexOf(
+    ['paypal', 'paywithgoogle', 'googlepay', 'amazonpay', 'applepay', 'klarna', 'klarna_account', 'klarna_paynow'].indexOf(
```
amihajlovski, line 76: I think you can only add klarna here since it uses indexOf, no need for klarna_account and klarna_paynow.

amihajlovski (dismissed):

amihajlovski (approved):

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/helpers.js
```diff
@@ -72,8 +72,19 @@ function displaySelectedMethod(type) {
     : type;
   resetPaymentMethod();
 
+  const klarnaVariants = ['klarna', 'klarna_account', 'klarna_paynow'];
```
maassenbas, line 75: Since we do `indexOf()`, then only `klarna` is enough

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/checkoutConfiguration.js
```diff
@@ -310,6 +310,27 @@ function getApplePayConfig() {
   };
 }
 
+function getKlarnaConfig() {
+  const {klarnaWidgetEnabled} = window;
```
maassenbas, line 314: `{ klarnaWidgetEnabled }`

maassenbas: Do we have this in ESLint?

amihajlovski: This will be fixed in another PR since the branch name is also not correct.

PR #921, opened 2023-06-12
Same merchant reference for all partial payments

zenit2001:
This PR implements the logic to use the same merchant reference for all partial payments.

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/checkBalance.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/checkBalance.js
@@ -1,5 +1,7 @@
 const BasketMgr = require('dw/order/BasketMgr');
 const Money = require('dw/value/Money');
+const OrderMgr = require('dw/order/OrderMgr');
+const Transaction = require('dw/system/Transaction');
 const AdyenHelper = require('*/cartridge/scripts/util/adyenHelper');
 const adyenCheckout = require('*/cartridge/scripts/adyenCheckout');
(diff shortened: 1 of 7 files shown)


src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/checkBalance.js
```diff
@@ -27,9 +29,14 @@ function getFormattedProperties(checkBalanceResponse, orderAmount) {
   return {};
 }
 
+// eslint-disable-next-line complexity
 function callCheckBalance(req, res, next) {
   try {
     const currentBasket = BasketMgr.getCurrentBasket();
+    let orderNo = currentBasket.custom?.adyenGiftCardsOrderNo;
```
zaiddreakh, line 36:
```suggestion
    let orderNo = currentBasket.custom?.adyenGiftCardsOrderNo || OrderMgr.createOrderNo();
```

src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/checkBalance.js
```diff
@@ -27,9 +29,14 @@ function getFormattedProperties(checkBalanceResponse, orderAmount) {
   return {};
 }
 
+// eslint-disable-next-line complexity
```
zaiddreakh, line 32: would be better if the function can be simplified instead of disabling the complexity rule

zaiddreakh (approved):

shanikantsingh (approved):

PR #924, opened 2023-06-26
Update the HMAC Key field to be empty instead of null

zenit2001:
This PR sets the HMAC Key field in BM to empty instead of null if no input is provided and updates the hmacVerification to not be false in case no hmac key is setup

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/notify.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/notify.js
@@ -7,18 +7,18 @@ const AdyenConfigs = require('*/cartridge/scripts/util/adyenConfigs');
  * Called by Adyen to update status of payments. It should always display [accepted] when finished.
  */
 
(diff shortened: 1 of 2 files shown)


amihajlovski (approved):

shanikantsingh (approved):

src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/adyen/notify.js
```diff
@@ -8,10 +8,10 @@ const AdyenConfigs = require('*/cartridge/scripts/util/adyenConfigs');
  */
 
 function handleHmacVerification(hmacKey, req) {
-  if (!hmacKey) {
-    return false;
+  if (hmacKey) {
+    return checkAuth.validateHmacSignature(req);
   }
-  return checkAuth.validateHmacSignature(req);
+  return true;
```
maassenbas, line 14: How are we handling cases without HMAC validation? 
Should we add a log to make a merchant aware that they are not using HMAC and that we are recommending using that?

PR #953, opened 2023-08-15
Payment instrument check

zenit2001:
This PR adds an additional check if the payment instruments exists before handling adyen giving component. In this way, if a merchant is using multiple payment service providers the order confirmation page will not fail rendering.

### DIFF
--- a/src/cartridges/int_adyen_overlay/cartridge/scripts/util/adyenHelper.js
+++ b/src/cartridges/int_adyen_overlay/cartridge/scripts/util/adyenHelper.js
@@ -156,6 +156,11 @@ var adyenHelperObj = {
   },
 
   getAdyenGivingConfig(order) {
+    if (!order.getPaymentInstruments(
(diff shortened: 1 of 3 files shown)


amihajlovski (approved):

src/cartridges/int_adyen_overlay/cartridge/scripts/util/adyenHelper.js
```diff
@@ -156,6 +156,11 @@ var adyenHelperObj = {
   },
 
   getAdyenGivingConfig(order) {
```
shanikantsingh, line 158: if `isAdyenGivingAvailable` is not used anywhere then it could be cleaned up

zenit2001: It is being used on line 169

shanikantsingh (approved):

PR #954, opened 2023-08-15
Show applepay button only safari

amihajlovski:
For the express option we can use the `isAvailable` property that checkout offers, but not at the end of checkout. 
At the end of checkout we have to use regex and not render the payment method.


zenit2001 (dismissed):

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/applePayExpress.js
```diff
@@ -123,182 +122,177 @@ function callPaymentFromComponent(data, resolveApplePay, rejectApplePay) {
...
+            paymentType: 'express',
+          };
+
+          const resolveApplePay = () => {
+            const finalPriceUpdate = {
               newTotal: {
                 type: 'final',
                 label: applePayConfig.merchantName,
-                amount: newCalculation.grandTotalAmount.value,
+                amount: `${applePayButtonConfig.amount.value / 100}`,
```
maassenbas, line 170: If it is a foreign currency with 1 or 3 decimals, then this might not be correct. Should we have a check here?

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/renderPaymentMethod.js
```diff
@@ -2,6 +2,8 @@ const store = require('../../../../store');
 const helpers = require('./helpers');
 const constants = require('../constants');
 
+const isSafari = /^((?!chrome|android|ios).)*safari/i.test(navigator.userAgent);
```
shanikantsingh, line 5: we could avoid using `isSafari` flag here. We could use the `isAvailable` function from checkout.
we could call line 176 `handlePayment(options);` before line 165 and then:
`const node = store.componentsObj[paymentMethodID]?.node;`
 `if (node.isAvailable) { `
`const paymentMethodAvailable = await node.isAvailable()`
`if (!paymentMethodAvailable) {`
`delete store.componentsObj[paymentMethodID]`
`return`
`}`
`}`

amihajlovski: I added it in another place, I had to move the code around in new functions because of big complexity.
Please check

src/cartridges/int_adyen_SFRA/cartridge/client/default/js/adyen_checkout/renderGenericComponent.js
```diff
@@ -222,25 +225,30 @@ async function initializeCheckout() {
   setAmazonPayConfig(store.checkout.paymentMethodsResponse);
   document.querySelector('#paymentMethodsList').innerHTML = '';
 
-  renderStoredPaymentMethods(
-    store.checkout.paymentMethodsResponse,
-    session.imagePath,
-  );
+  const paymentMethodsWithoutGiftCards =
```
maassenbas, line 228: Instead of filtering the gift cards, can we also add `giftcard` as blocked type in the configuration?

shanikantsingh (approved):

zenit2001 (approved):

shanikantsingh (approved):

PR #969, opened 2023-09-22
Troubleshooting section in config page

zenit2001:
This PR adds a troubleshooting section to the custom BM.


src/cartridges/bm_adyen/cartridge/static/default/js/adyenSettings.js
```diff
@@ -171,6 +176,39 @@ document.addEventListener('DOMContentLoaded', () => {
     window.open(window.classicConfigPageUrl);
   }
 
+  function downloadFile(filePath) {
+    const link = document.createElement('a');
+    link.href = filePath;
+    link.download = filePath.substr(filePath.lastIndexOf('/') + 1);
+    link.click();
+  }
```
amihajlovski, line 184: I see you only add html elements but never remove it. Can we do that?

amihajlovski (approved):

shanikantsingh (approved):

PR #984, opened 2023-10-18
fix: blockedPaymentMethods referencing the removed constants

rge-dS:
## Summary
This change fixes the undefined blockedPaymentMethods because the deleted constants from AdyenHelper were still being used. 
I used adyenSessions as a reference, in which this was already fixed.


amihajlovski (approved):

zenit2001 (approved):

amihajlovski:
Thanks a lot for the contribution @rge-dS 
We will include this in the next release!

Kind Regards
Aleksandar
Adyen

PR #1025, opened 2023-12-06
[SFI-556] Level 2/3 data only on scheme requests

zenit2001:
## Summary
Level 2/3 data is being included in all payment requests in case it is enabled.
This PR includes the level 2/3 data only in case `scheme` payment methods are being used.

## Tested scenarios
- Scheme payment with level 2/3 data enabled
- Alternative payment method with level 2/3 data enabled
- Scheme payment with level 2/3 data disabled

**Fixed issue**:  SFI-556

### DIFF
--- a/src/cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.js
+++ b/src/cartridges/int_adyen_overlay/cartridge/scripts/adyenCheckout.js
@@ -49,6 +49,7 @@ function createPaymentRequest(args) {
     );
 
     paymentRequest = AdyenHelper.add3DS2Data(paymentRequest);
+    const paymentMethodType = paymentRequest.paymentMethod.type;
 
     // Add Risk data
     if (AdyenConfigs.getAdyenBasketFieldsEnabled()) {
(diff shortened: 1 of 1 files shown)


amihajlovski (approved):

shanikantsingh (commented):
should this fix be also applied in [adyenSessions.js](https://github.com/Adyen/adyen-salesforce-commerce-cloud/blob/d57438f90c3b452b2324f2f96571b3e27f68727d/src/cartridges/int_adyen_overlay/cartridge/scripts/adyenSessions.js#L64)?

zenit2001:
> should this fix be also applied in [adyenSessions.js](https://github.com/Adyen/adyen-salesforce-commerce-cloud/blob/d57438f90c3b452b2324f2f96571b3e27f68727d/src/cartridges/int_adyen_overlay/cartridge/scripts/adyenSessions.js#L64)?

In `adyenSessions.js`, we don't have the payment method selected, since we use it only for rendering the payment methods and not for sending the payment request.

shanikantsingh (approved):

PR #1032, opened 2024-01-09
[SFI-516] Different live prefixes can be used accross sites

zenit2001:
## Summary
Enabling the usage of different LIVE prefixes between different sites.
Possibility to use different LIVE prefixes, based on region merchant site is located.

## Tested scenarios
- iDeal Payment
- Apple Pay Payment

**Fixed issue**:  SFI-516

### DIFF
--- a/src/cartridges/bm_adyen/cartridge/static/default/js/adyenSettings.js
+++ b/src/cartridges/bm_adyen/cartridge/static/default/js/adyenSettings.js
@@ -473,7 +473,7 @@ document.addEventListener('DOMContentLoaded', () => {
   });
 
   testRadio.addEventListener('click', () => {
-    livePrefix.disabled = testRadio.value === 'TEST';
+    livePrefix.disabled = true;
   });
 
   adyenGivingBackground.addEventListener('click', saveAndHideAlerts);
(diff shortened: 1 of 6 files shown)


src/cartridges/bm_adyen/cartridge/static/default/js/adyenSettings.js
```diff
@@ -465,6 +468,14 @@ document.addEventListener('DOMContentLoaded', () => {
     }
   });
 
+  productionRadio.addEventListener('click', () => {
+    livePrefix.disabled = productionRadio.value === 'TEST';
```
amihajlovski, line 472: this should be 
`livePrefix.disabled = false` 
right now it works because productionRadio.value will always be LIVE ;)

shanikantsingh (dismissed):

src/cartridges/bm_adyen/cartridge/static/default/js/adyenSettings.js
```diff
@@ -465,6 +468,14 @@ document.addEventListener('DOMContentLoaded', () => {
     }
   });
 
+  productionRadio.addEventListener('click', () => {
+    livePrefix.disabled = false;
+  });
+
+  testRadio.addEventListener('click', () => {
+    livePrefix.disabled = testRadio.value === 'TEST';
```
amihajlovski, line 476: this should always be `true`
```
livePrefix.disabled = true;
```

zenit2001: testRadio.value will always be TEST, however changed it to be true.

shanikantsingh (approved):

PR #1075, opened 2024-04-25
Enabling github actions for SFRA 7

shanikantsingh:
## Summary
- run e2e tests for SFRA7
- run e2e tests for SFRA7

## Tested scenarios

**Fixed issue**:  #SFI-698

### DIFF
--- a/tests/playwright/fixtures/BRL.spec.mjs
+++ b/tests/playwright/fixtures/BRL.spec.mjs
@@ -24,7 +24,7 @@ for (const environment of environments) {
 
       await checkoutPage.goToCheckoutPageWithFullCart(regionsEnum.BR);
       await checkoutPage.setShopperDetails(shopperData.BR);
-      if (environment.name.indexOf("v6") === -1) {
+      if (environment.name.indexOf('v5') !== -1) {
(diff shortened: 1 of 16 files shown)


tests/playwright/fixtures/BRL.spec.mjs
```diff
@@ -24,7 +24,7 @@ for (const environment of environments) {
 
       await checkoutPage.goToCheckoutPageWithFullCart(regionsEnum.BR);
       await checkoutPage.setShopperDetails(shopperData.BR);
-      if (environment.name.indexOf("v6") === -1) {
+      if (environment.name.indexOf('v5') !== -1) {
```
amihajlovski, line 27: what is the logic behind this?
because previous logic was to set the email if the environment is not v6.
no with this logic it will set email also for v6 and v7.
shouldn't it stay only for v5?

shanikantsingh: earlier it was setting email if not v6. which would have set it for v5 and v7. Now it checks for v5 and sets email only for v5.

amihajlovski (approved):

zenit2001 (approved):

PR #1120, opened 2024-07-16
Adding metadata upload workflow

zenit2001:
## Summary
Automating the metadata import step.
No need to upload metadata manually in case new fields are added.

## Tested scenarios
- Automatic metadata field import in case of changes in /metadata folder
- In case no changes in the /metadata folder, the pipeline will not run

**Fixed issue**: SFI-58


dcardos (approved):
All looks good, I assume the timeout of 5 minutes is enough to run the job most of the times.

PR #1122, opened 2024-07-18
PayPal Express in Cart and Mini-Cart page

zenit2001:
This PR enables the PayPal express as a payment method in SFCC cartridge, including the possibility to use the flow with order review page and without order review page. 
Enablement of PayPal Express

## Tested scenarios
- PayPal express payments with order review page
- PayPal express payments without order review page
- Shipping method changes
- Address changes

**Fixed issue**:  SFI-42


src/cartridges/app_adyen_SFRA/cartridge/client/default/js/paypalExpress.js
```diff
@@ -0,0 +1,225 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(data),
+    });
+    const { action, errorMessage = '' } = await response.json();
+    if (response.ok && action) {
+      component.handleAction(action);
+    } else {
+      throw new Error(errorMessage);
```
dcardos, line 23: I guess we are assuming an error message will be populated, but would't be better for debug something like: 
`throw new Error(errorMessage || 'Express payment action failed');`

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/paypalExpress.js
```diff
@@ -0,0 +1,225 @@
...
+    }
+  } catch (e) {
+    component.handleError();
```
dcardos, line 26: Are we ignoring the error message here?

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/paypalExpress.js
```diff
@@ -0,0 +1,225 @@
...
+  redirect.submit();
+}
+
+function makeExpressPaymentDetailsCall(data) {
+  return $.ajax({
+    type: 'POST',
+    url: window.makeExpressPaymentDetailsCall,
+    data: JSON.stringify({ data }),
+    contentType: 'application/json; charset=utf-8',
+    async: false,
```
dcardos, line 67: It seems this is deprecated. Are we sure we want to use async:false here? [Link](https://stackoverflow.com/questions/11448011/jquery-ajax-methods-async-option-deprecated-what-now) for more info.

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/paypalExpress.js
dcardos: This applies to all Error handling in the code.

src/cartridges/int_adyen_SFRA/cartridge/adyen/scripts/expressPayments/paypal/makeExpressPaymentDetailsCall.js
```diff
@@ -0,0 +1,69 @@
...
+  try {
+    const request = JSON.parse(req.body);
+    const currentBasket = BasketMgr.getCurrentBasket();
+
+    const response = adyenCheckout.doPaymentsDetailsCall(request.data);
+
+    paypalHelper.setBillingAndShippingAddress(currentBasket);
+
+    // Setting the session variable to null after assigning the shopper data to basket level
+    session.privacy.shopperDetails = null;
```
dcardos, line 38: Is this okay in the sense it will not affect other parts of the application unintentionally?

dcardos (commented):
Great work, a few comments that might not require code changes, but it would be nice to double check.

src/cartridges/int_adyen_SFRA/cartridge/adyen/scripts/expressPayments/paypal/makeExpressPaymentDetailsCall.js
zenit2001: This is used only for PayPal express and not for other payment method. Once we assign the data in basket level, we need to clean

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/paypalExpress.js
zenit2001: This is a front-end part, so ideally we want to keep the logging as limited as possible there. If there is an error coming from the response we will throw it, otherwise we leave it empty.

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/paypalExpress.js
zenit2001: Yeah, we ignore it also for other payment methods. Ideally, we don't want to log on the front-end

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/paypalExpress.js
zenit2001: You're right, it is deprecated, but there is not any other official alternative provided by jquery. There exists only manual solutions. 
We have discussed this as part of unifying all the calls we do, however we need to come up with a generic alternative as we use it in quite some calls in the cartridge.

dcardos (approved):

dcardos (approved):

amihajlovski (approved):

PR #1173, opened 2024-09-17
fix problem with payment method value in custom attribute

AVATOR:
## Summary
I check if the value is an object before using JSON.stringify. The issue arises because result.paymentMethod.type is a string, and after converting it with JSON.stringify, the value becomes "paypal", which breaks the OMS and ERP systems.
image

## Tested scenarios

**Fixed issue**:


zenit2001:
Hi @AVATOR ,

Thanks for creating the PR.
Can you please let us know if you are using our [OMS](https://github.com/Adyen/adyen-salesforce-oms) or some third party developed OMS?

Kind regards,
Zenit 
Adyen

AVATOR:
No. We don't use your OMS code. We have code that was implemented before. OMS grabe order without problems but ERP skip payment information if value `"paypal"` instead of `paypal`

zenit2001:
Hi @AVATOR ,

As this change is related to your ERP system, you can keep the change in your codebase, but we are not integrating it in our cartridge as it does not benefit our use cases.
Hence, I'm closing this PR.

Kind regards,
Zenit
Adyen

PR #1228, opened 2024-12-16
Feature/adyen error objects

shanikantsingh:
## Summary
- Refactor try catch block
- Create custom Adyen error type
- Refactor try catch block
- Create custom Adyen error type

## Tested scenarios
- card payment flow

**Fixed issue**:  #SFI-1007 #SFI-1025

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/adyen/scripts/donations/adyenGiving.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/adyen/scripts/donations/adyenGiving.js
@@ -31,81 +31,75 @@ const { AdyenError } = require('*/cartridge/adyen/logs/adyenError');
 
 // eslint-disable-next-line complexity
 function donate(donationReference, donationAmount, orderToken) {
-  try {
-    if (session.privacy.orderNo !== donationReference) {
-      throw new AdyenError('Donation reference is invalid');
-    }
-
-    let paymentMethodVariant;
-    const order = OrderMgr.getOrder(donationReference, orderToken);
-    const paymentInstrument = order.getPaymentInstruments(
-      AdyenHelper.getOrderMainPaymentInstrumentType(order),
-    )[0];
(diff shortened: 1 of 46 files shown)


src/cartridges/int_adyen_SFRA/cartridge/adyen/scripts/donations/adyenGiving.js
```diff
@@ -26,12 +26,14 @@ const AdyenHelper = require('*/cartridge/adyen/utils/adyenHelper');
 const AdyenConfigs = require('*/cartridge/adyen/utils/adyenConfigs');
 const constants = require('*/cartridge/adyen/config/constants');
 const AdyenLogs = require('*/cartridge/adyen/logs/adyenCustomLogs');
+const setErrorType = require('*/cartridge/adyen/logs/setErrorType');
+const { AdyenError } = require('*/cartridge/adyen/logs/adyenError');
 
 // eslint-disable-next-line complexity
 function donate(donationReference, donationAmount, orderToken) {
   try {
```
zenit2001, line 34: Should we remove try/catch here, considering we are adding it at main function `donation()`?

zenit2001 (commented):
Looks good, but we would need to consider the points below also, to ensure consistency: 
- None of adyenDeleteRecurringPayment and deletePayment has a try/catch logic in them, meaning that if an error is thrown during deleting a stored payment method, we will not get any insight. I think we should apply the logic there also.
- zeroAuthPayment is inside a try catch , however it’s a helper function being used by savePayment(). Should we move try/catch logic in savePayment() instead?
- Should we consider adding an AdyenError in updateSavedCards()?
- Should we consider adding try/catch to notify() and remove try/catch on helper functions related to that. Instead we can replace them with AdyenError?
- Same as the above applies to begin.js file, removing the try/catch from the helpers and adding it to the main function.

zenit2001 (approved):

PR #1255, opened 2025-02-11
Moving the place order logic into webhook processing part

zenit2001:
## Summary

## Tested scenarios

**Fixed issue**:


src/cartridges/int_adyen_SFRA/cartridge/adyen/webhooks/handleCustomObject.js
```diff
@@ -152,6 +158,7 @@ function handle(customObj) {
               `Partial amount ${customObj.custom.value} received for order number ${order.orderNo} with total amount ${totalAmount}`,
             );
           } else {
+	    placeOrder(order);
             order.setPaymentStatus(Order.PAYMENT_STATUS_PAID);
             order.setExportStatus(Order.EXPORT_STATUS_READY);
```
amihajlovski, line 163: In the placeOrder function you are only setting a log in case of an error, and the rest continues meaning the payment status will be set to PAID. 
Throw an error in the function so it can be caught in the catch block and the payment and order status is not updated.

shanikantsingh (approved):

amihajlovski (approved):

PR #1268, opened 2025-02-20
Feature/sfi 974 remove sfra version dep

shanikantsingh:
## Summary
- remove SFRA5 6 and 7 dependency.
- - remove SFRA5 6 and 7 dependency.

## Tested scenarios
3DS and non-3DS flow

**Fixed issue**:


amihajlovski (approved):

src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/checkout_services/adyenCheckoutServices.js
```diff
@@ -3,7 +3,7 @@ const collections = require('*/cartridge/scripts/util/collections');
 
 function processPayment(order, handlePaymentResult, req, res, emit) {
   res.json({
-    error: false,
+    error: true,
```
zenit2001, line 6: Any reason on setting `error: true` here? If not, can we revert back to how it was?

shanikantsingh: For [SFRA6](https://github.com/SalesforceCommerceCloud/storefront-reference-architecture/blob/bef36e6b2efc0680da5c6aedf060fb9baba5ea13/cartridges/app_storefront_base/cartridge/client/default/js/checkout/checkout.js#L367) and [SFRA5](https://github.com/SalesforceCommerceCloud/storefront-reference-architecture/blob/50ee82face6e0a000f649a51f162e8a3f171531c/cartridges/app_storefront_base/cartridge/client/default/js/checkout/checkout.js#L334) in checkout.js contiueUrl is needed for else part of code.

if its not passed from backend then in SFRA5 it throughs error as it tries to do indexOf of undefined 

By setting error to true the control will not go to else part when there is adyenAction

I didn't want to modify the checkout.js file within our cartridge for this fix.

PR #1316, opened 2025-04-22
e2e reliability workflow

shanikantsingh:
## Summary
- setup e2e reliability workflow
- setup e2e reliability workflow

## Tested scenarios

**Fixed issue**:


zenit2001 (commented):
Instead of duplicating a current job that we have (`E2E_SFRA.yml`), could we possibly change the existing workflow only for this branch to satisfy both conditions ( running on a pull request and also on a schedule )?

Something like below would satisfy these conditions I think
on:
    schedule:
         - cron : .....
    pull_request:

amihajlovski (approved):

zenit2001 (approved):

PR #1317, opened 2025-04-24
Setting payment instrument fields for paymentFromComponent based on stateData

zenit2001:
## Summary
Payment instrument fields for payment method are set based on state data from the component same as they are passed to `/payments` call
Removing the possibility of the fields to be saved as HTML translated tags.

## Tested scenarios
- Card payments
- PaymentFromComponent payments
- Zero auth payments
- Giftcard payments
- Express payments

**Fixed issue**:  SFI-1170

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/adyen/utils/adyenHelper.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/adyen/utils/adyenHelper.js
@@ -31,7 +31,7 @@ const StringUtils = require('dw/util/StringUtils');
 const Money = require('dw/value/Money');
 const BasketMgr = require('dw/order/BasketMgr');
 const OrderMgr = require('dw/order/OrderMgr');
-//script includes
+// script includes
 const ShippingMethodModel = require('*/cartridge/models/shipping/shippingMethod');
 const collections = require('*/cartridge/scripts/util/collections');
 const constants = require('*/cartridge/adyen/config/constants');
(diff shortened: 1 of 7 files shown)


amihajlovski (approved):

src/cartridges/int_adyen_SFRA/cartridge/adyen/utils/adyenHelper.js
```diff
@@ -537,20 +563,53 @@ let adyenHelperObj = {
...
+    paymentInstrument,
+    customerEmail,
+  ) {
     const jsonObject = JSON.parse(paymentInstrument.custom.adyenPaymentData);
 
     const filteredJson = adyenHelperObj.validateStateData(jsonObject);
     const { stateData } = filteredJson;
 
+    Transaction.wrap(() => {
+      adyenHelperObj.setPaymentInstrumentFields(paymentInstrument, stateData);
```
shanikantsingh, line 602: Could we move setPaymentInstrumentFields function call before or after `createAdyenRequestObject` function is called. The reason is `createAdyenRequestObject ` being a helper function should have only one responsibility of returning a request object. To avoid any side effects it would be good to move `setPaymentInstrumentFields` function call a level up.

zenit2001: The reasons of using `createAdyenRequestObject()` for setting these fields are:
- We use the validated state data which are also passed to the /payments call
- We have access to `paymentInstrument`

Moving it one step before it's not possible, since the state data will not be present, while moving it after leads to the same issue we had; having translated fields as the data is retrieved from the form.

The state data should return the payment method data in a formatted and generic way, however, in case that doesn't happen, we can use a try/catch block in `setPaymentInstrumentFields()` and throw an error log in case of any inconsistency.

amihajlovski (approved):

shanikantsingh (approved):

PR #1365, opened 2025-06-25
[SFI-1243] Initialize Fastlane and render watermark

amihajlovski:
## Summary
Initialize fastlane component on guest checkout page
Being able to authenticate user

## Tested scenarios
- Fastlane flow

**Fixed issue**:  SFI-1243

### DIFF
--- a/src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/checkout/index.js
+++ b/src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/checkout/index.js
@@ -256,7 +256,7 @@ function init() {
       const guestEmail = document.querySelector('#email-guest');
       if (guestEmail) {
         const watermarkDiv = document.createElement('div');
-        watermarkDiv.id = '#watermark-container';
+        watermarkDiv.id = 'watermark-container';
         guestEmail.parentElement.appendChild(watermarkDiv);
         const fastlane = await window.AdyenWeb.initializeFastlane(
           store.checkoutConfiguration,
--- a/src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/checkout/index.js
+++ b/src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/checkout/index.js
@@ -248,6 +248,22 @@ function registerFirstPaymentMethod() {
 function init() {
   $(document).ready(async () => {
     paymentMethodsResponse = await getPaymentMethods();
+    const isFastlaneEnabled =
+      paymentMethodsResponse?.AdyenPaymentMethods?.paymentMethods.some(
+        (pm) => pm.type === 'fastlane',
+      );
+    if (isFastlaneEnabled) {
+      const guestEmail = document.querySelector('#email-guest');
+      if (guestEmail) {
+        const watermarkDiv = document.createElement('div');
+        watermarkDiv.id = '#watermark-container';
+        guestEmail.parentElement.appendChild(watermarkDiv);
+        const fastlane = await window.AdyenWeb.initializeFastlane(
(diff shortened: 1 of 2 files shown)


zenit2001 (approved):

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/checkout/index.js
```diff
@@ -247,14 +247,37 @@ function registerFirstPaymentMethod() {
...
+    try {
+      paymentMethodsResponse = await getPaymentMethods();
+      const paymentMethods =
+        paymentMethodsResponse?.AdyenPaymentMethods?.paymentMethods;
+      const isFastlaneEnabled = paymentMethods?.some(
+        (pm) => pm.type === 'fastlane',
+      );
+      if (isFastlaneEnabled) {
+        const guestEmail = document.querySelector('#email-guest');
+        if (guestEmail) {
```
shanikantsingh, line 261: Here the only way we are identifying the guest shopper flow is using the "guestEmail". Should we also check if the shopper is logged in or not to determine the guest shopper flow?

shanikantsingh (dismissed):

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/checkout/index.js
```diff
@@ -247,14 +247,37 @@ function registerFirstPaymentMethod() {
...
+          watermarkContainer = document.createElement('div');
+          watermarkContainer.id = 'watermark-container';
+          guestEmail.parentElement.appendChild(watermarkContainer);
+          fastlane = await window.AdyenWeb.initializeFastlane(
```
shanikantsingh, line 265: fastlane can be a "const" as it is not used outside the block scope. Also watermarkContainer could be a const.
In the Catch block we should query the element, check that the element exist before calling remove function.

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/checkout/index.js
```diff
@@ -247,14 +247,37 @@ function registerFirstPaymentMethod() {
...
+      }
+      const storedCustomerEmail = sessionStorage.getItem('customerEmail');
+      $('body').trigger('checkout:renderPaymentMethod', {
+        email: storedCustomerEmail,
+      });
+      helpers.createShowConfirmationForm(
+        window.ShowConfirmationPaymentFromComponent,
+      );
+    } catch (err) {
+      watermarkContainer.remove();
```
shanikantsingh, line 279: In the current situation if the getPaymentMethods function call at line 253 throws an error then in catch block we would have error as element "watermarkContainer" would be undefined.

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/checkout/index.js
amihajlovski: I moved the logic to the backend handler.

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/checkout/index.js
amihajlovski: Done.

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/checkout/index.js
amihajlovski: This is also handled.

shanikantsingh (approved):

PR #1366, opened 2025-07-03
[SFI-150] Pre auth hook

zenit2001:
## Summary
Implementing an empty handler to allow pre-auth checks before sending the /payments call to Adyen.
It allows merchants to do pre-authorization checks before authorizing a payment.

## Tested scenarios
- Pre-auth returning error:false
- Pre-auth returning error:true

**Fixed issue**:  SFI-150

### DIFF
--- a/src/cartridges/int_adyen_SFRA/cartridge/adyen/scripts/payments/adyenCheckout.js
+++ b/src/cartridges/int_adyen_SFRA/cartridge/adyen/scripts/payments/adyenCheckout.js
@@ -57,6 +57,16 @@ function doPaymentsCall(order, paymentInstrument, paymentRequest) {
   } else if (transactionAmount !== paymentRequest?.amount?.value) {
     throw new AdyenError('Amounts dont match');
   }
+  // Pre-auth hook
+  const preAuthResult = hooksHelper(
+    'app.payment.pre.auth',
+    'preAuthorization',
+    paymentRequest,
+    preAuthorizationHook.preAuthorization,
+  );
+  if (preAuthResult?.error) {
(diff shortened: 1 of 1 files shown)


amihajlovski (approved):

shanikantsingh (changes requested):
We could implement the feature using HookMgr rather than using hooksHelper. Using HookMgr the hook for pre-auth could be implemented something like:
```
if (HookMgr.hasHook('app.payment.pre.auth.processor.' + processor.ID.toLowerCase())) {
                preAuthResult = HookMgr.callHook('app.payment.pre.auth.processor.' + processor.ID.toLowerCase(),
                    'Handle',
                    order,
                    paymentRequest
                );
            }
              if (preAuthResult?.error) {
                throw new AdyenError('Pre Auth check failed', preAuthResult);
              }
```
This approach ties the pre-auth hook with the payment processor. So merchant would be able to implement different pre-auth hook for different  payment processors.
Also, it would be good to make the hooks call just before calling AdyenHelper.executeCall function because there are already certain validations happening in doPaymentsCall. I think we should call pre-auth hook only if those validations pass.
Also we should introduce the pre-auth check hook for partial payments and zero auth flows. It would be beneficial to understand and document the pre-auth check for pos flow too. (This suggestion should be discussed with product as it may cause a scope creep)
With this implementation we do not need to implement preAuthorizationHandling.js file and also we don not need to add the hook in hooks.json. It would be up to the merchant to do it if they are implementing pre-auth hook.
We should add them as part of documentation only.
Let me know your thoughts on this approach.

PR #1434, opened 2025-11-11
[SFI-1439] Failure scenario not properly handled for giftcards and payments without action

zenit2001:
## Summary
Fixing the failure scenario for giftcards in combination with a no 3ds card.
Shopper is able to fix the wrong card data and continue paying with giftcards

## Tested scenarios
- Giftcards + no 3DS cards

**Fixed issue**:  SFI-1439


src/cartridges/int_adyen_SFRA/cartridge/controllers/middlewares/checkout_services/placeOrder.js
```diff
@@ -163,9 +165,25 @@ function placeOrder(req, res, next) {
...
+            try {
+                cancelPartialPaymentOrderHelper(currentBasket);
+                basketDataCleared = true; 
+            } catch (error) {
+                AdyenLogs.error_log('Failed to cancel partial payment order on payment failure:', error);
+            }
+        }
+        Transaction.wrap(function () { OrderMgr.failOrder(order, true); });
+        // Clear basket data if not already cleared by the helper
+        if (!basketDataCleared) {
```
shanikantsingh, line 180: The "basketDataCleared" flag could be removed if `clearAdyenBasketData` function is not called in cancelPartialPaymentOrderHelper. We could call the clearAdyenBasketData` function directly here without the if condition.

shanikantsingh (approved):

PR #1438, opened 2025-11-25
[SFI-1422] Apple pay ( and other pm ) use isAvailable() while rendering if it exists

zenit2001:
## Summary
Apple Pay renders on unsupported devices/browsers. 
It uses `isAvailable()` property while rendering payment methods.

## Tested scenarios
- Apple Pay on renders on iOS ( Safari, Chrome )
- Apple Pay doens't render on Android

**Fixed issue**: SFI-1422

### DIFF
--- a/src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/express/paymentMethods/applepay/applepay.js
+++ b/src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/express/paymentMethods/applepay/applepay.js
@@ -296,6 +296,18 @@ class ApplePay {
     };
   }
 
+  static async checkIfComponentIsAvailable(component) {
+    if (typeof component.isAvailable === 'function') {
+      try {
+        const isComponentAvailable = await component.isAvailable();
+        return isComponentAvailable;
+      } catch (error) {
+        return false;
+      }
+    }
+    return true;
+  }
+
   async getComponent() {
     const checkout = await initializeCheckout(
       this.applicationInfo,
@@ -307,10 +319,12 @@ class ApplePay {
       checkout,
       applePayConfig,
     );
-    if (!component.isAvailable()) {
-      return null;
+    const isComponentAvailable =
+      await ApplePay.checkIfComponentIsAvailable(component);
+    if (isComponentAvailable !== false) {
+      return component;
     }
-    return component;
+    return null;
   }
 }
 


src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/checkout/renderPaymentMethod.js
```diff
@@ -40,7 +40,19 @@ const getComponentConfig = (paymentMethodID, paymentMethod) => {
   return { ...baseConfig, ...additionalConfig };
 };
 
-function setNode(paymentMethod, paymentMethodID) {
+async function checkIfNodeIsAvailable(node) {
```
shanikantsingh, line 43: The checkIfNodeIsAvailable function now return either boolean `false` or `undefined`. We could also return `true` from the function which would make the intent of the function more clear.
```suggestion
async function checkIfNodeIsAvailable(node) {
// if the isAvailable is not present then node is considered as available
  if (typeof node?.isAvailable !== 'function') {
    return true;
  }

  try {
// if the isAvailable function resolves then node is considered as available
    await node.isAvailable();
    return true;
  } catch (error) {
    return false;
  }
}
```

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/checkout/renderPaymentMethod.js
```diff
@@ -146,6 +164,19 @@ function createListItem(paymentMethodID, liContents) {
   return li;
 }
 
+function mountComponentIfAvailable(paymentMethodID, container, li) {
+  const componentObj = store.componentsObj[paymentMethodID];
+  const isAvailable = componentObj?.isAvailable;
+
+  if (isAvailable !== false) {
```
shanikantsingh, line 171: Based on above suggestion this can check for isAvailable to be true.

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/express/paymentMethods/applepay/applepay.js
```diff
@@ -302,7 +302,15 @@ class ApplePay {
       this.translations,
     );
     const applePayConfig = this.getConfig();
-    return window.AdyenWeb.createComponent(APPLE_PAY, checkout, applePayConfig);
+    const component = window.AdyenWeb.createComponent(
+      APPLE_PAY,
+      checkout,
+      applePayConfig,
+    );
+    if (!component.isAvailable()) {
```
shanikantsingh: `isAvailable` is an async function so it should either be used with `await` or `then...catch`

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/checkout/renderPaymentMethod.js
zenit2001: The function returns `true` in two use cases ( check line 46,52 ).
This is not possible in my opinion, since `isAvailable()` callback is not available for all types of components and most of the times returns undefined, while the `renderPaymentMethod()` is used for the whole rendering in the end of checkout. 
Considering the above, `checkIfNodeIsAvailable()` 
- returns true when `isAvailable()` callback doesn't exists in the component, such that rendering is not blocked
- returns the value also when `isAvailable()` callback exists in the component (e.g. applepay)
- returns false otherwise

This way, every time `checkIfNodeIsAvailable()` returns something different from false, it means that we are safe to render the component.

shanikantsingh: Yes, the `isAvailable()` function returns promise and is expected to return `Promise.resolve` when component is available otherwise it return Promise.reject() which would trigger the catch block of checkIfNodeIsAvailable function.

src/cartridges/app_adyen_SFRA/cartridge/client/default/js/adyen/express/paymentMethods/applepay/applepay.js
zenit2001: This is fixed in last commit

shanikantsingh (approved):

